Issue #8, opened 2024-05-16
Add GPU measurements to Benchmark Script


fabianlim:
We should include the gpu consumption into `FILE_RESULTS`
- new column `gpu_mem` (this can be in MiB)

The GPU memory can be collected by:
- starting a async process running `nvidia-smi` and log to a `FILE_MEM`
- perform the `subprocess.run`
- read `FILE_MEM` and process out how much memory was consumed

Points to note:
- As part of the processing we should perform some estimation based on the gpu memory consumed by the device before the `subprocess.run` and account for that appropriately. 
- If there are multiple GPUs used how should we report? should it be the average?
- We use `CUDA_VISIBLE_DEVICES` as environment to the `subprocess`, but Im not sure which GPU `nvidia-smi` will return. The best is to specifiy the device /s ids to `nvidia-smi` call. 
- we ise `nvidia-smi` which reports the `reserved_memory` not the `allocated_memory`. This needs to be documented.

PR #14, opened 2024-05-17
Provide Memory Benchmarking Feature to Benchmarking Code

achew010:
# Description
This PR adds GPU memory logging features to the benchmark script according to #8 and an updated benchmark README for usage instructions.

There are 2 approaches to logging memory,
- Using Nvidia's SMI CLI tool
- Using Huggingface's HFTrainer's API

Note: Issue #19 is created to address the grouping of memory values using a common prefix and will be addressed in future

# Usage

## 1. Nvidia's SMI CLI tool

Set environment variable `MEMORY_LOGGING=nvidia` to use `run_benchmarks.sh` with nvidia logging

For each experiment, 

- Before the experiment calls `subprocess.run`, it will open an async `nvidia-smi` process to monitor only gpu indices in `$CUDA_VISIBLE_DEVICES` and log to `FILE_MEM` inside `Experiment.save_dir`
- After the experiment `subprocess` call is completed. It terminates the async process.
- At the end of each experiment, reads and aggregates an average memory over time (MiB / per sec) for each device and finally saves the average memory across all devices into `gpu_mem` in the main result logging function `Experiment.write_results`
- Since it is an independent process called to measure the device, no expected slowdown in training speed

Each experiment directory will have a gpu log that contains ```<Timestamp>, <GPU Name>, <GPU ID>, <GPU Memory Used>```

The memory readings will be reflected in the results `raw_summary.csv` under the column `'nvidia_mem_reserved'` where the raw values are reported in `MiB`

## 2.  Torch CUDA through Huggingface's HFTrainer's API

Set environment variable `MEMORY_LOGGING=huggingface` to use `run_benchmarks.sh` with huggingface logging **(default)**

HFTrainer has a feature to log memory through the `skip_memory_metrics=False` training argument. In their [documentation](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.skip_memory_metrics), it is mentioned that setting this argument to `False` will affect training speed. In our tests so far (below), we do not see significant difference in throughput (tokens/sec) when using this argument.

A set of finegrain GPU readings will show as additional columns in the results `raw_summary.csv` where the raw values are reported in `bytes`

image

## 3. Log Both

Set environment variable `MEMORY_LOGGING=all` to use `run_benchmarks.sh` with both logging methods

### 1. The Nvidia-SMI Utility is a coarse measurement tool that captures anything takes up GPU memory. It is simple and non-intrusive as it doesn't involve probing the trainer. It uses the NVML library to fetch reserved memory for each device ID -

- All running computer processes using that device
- All gpu memory used inside the training script
- Any other caching of gpu memory by the training script

Note: To get accurate measurements, no other processes should be running on the device apart from the target process itself.

### 2. The HFTrainer API is a more precise tool that logs memory usage for a couple of operations inside HFTrainer
It uses `torch.cuda.memory_allocated` to probe the trainer by taking snapshots of allocated memory and storing the differences between the before and after of each stage. The following stages are probed -
- Before Trainer init
- `Trainer.__init__`, 
- `Trainer.train`, 
- `Trainer.evaluate`, 
- `Trainer.predict`. 

Note: Any gpu memory accessed and used outside any of these stages or not part of HFTrainer will not be tracked. If the train script does not use the Huggingface trainer then this API will not work as well.

### Note: Details on Memory Calculations from HFTrainer for GPTQ-LoRA + FSDP 

This is an example of the memory values that HFTrainer will produce in the outputs of `train()`
```
output_metrics = {
    'train_runtime': 191.2491, 
    'train_samples_per_second': 0.209, 
    'train_steps_per_second': 0.052, 
    'train_tokens_per_second': 428.342, 
    'train_loss': 1.0627506256103516, 
    'init_mem_cpu_alloc_delta': 4096, 
    'init_mem_gpu_alloc_delta': 0, 
    'init_mem_cpu_peaked_delta': 0, 
    'init_mem_gpu_peaked_delta': 0, 
    'train_mem_cpu_alloc_delta': 839086080, 
    'train_mem_gpu_alloc_delta': -17491768832, 
    'train_mem_cpu_peaked_delta': 0, 
    'train_mem_gpu_peaked_delta': 26747825664, 
    'before_init_mem_cpu': 5513297920, 
    'before_init_mem_gpu': 36141687296, 
    'epoch': 0.01
}
```

We refer to the keys of the memory metrics in this order 
 - `before_init_mem_X` as stage0 
 - `init_mem_X` as stage1 
 - `train_mem_X` as stage2
 - ... 

We currently compute the memory values in the report by taking the largest of sums. For example:

For allocated memory value
```
max([
  stage0_mem + stage1_allocated_delta, 
  stage0_mem + stage1_allocated_delta + stage2_allocated_delta,
  ...
])
```

For peak memory value
```
max([
  stage0_mem + stage1_allocated_delta + stage1_peaked_delta, 
  stage0_mem + stage1_allocated_delta + stage2_allocated_delta + stage2_peaked_delta,
  ...
])
```

Notice that we do not include `stage0_mem` alone when computing the max value. This is to avoid misleading comparisons between GPTQ-LoRA and other approaches that support low-memory mode. GPTQ-LoRA + FSDP currently does not support low-memory mode as mentioned #18. 

The `stage0_mem` value of GPTQ-LoRA + FSDP will reflect a larger  value as it is loaded fully before the trainer is initialized and then subsequently will be sharded internally in `trainer.prepare`. 

This might cause some misleading comparisons when other variants are loaded in low-memory mode and have smaller `stage0_mem` memory consumption than GPTQ-LoRA + FSDP before its sharding. Once low-memory mode is supported for GPTQ-LoRA, we will include `stage0_mem` back inside the max computation

# Tests

### 1. No Significant Slowdown From Using HFTrainer Memory Probes API on QLoRA Training
In both the Mistral7B model and Mixtral model, introducing the memory probes do not show a significant impact to the throughput of the training run (50 steps). Generally with larger batch sizes and models, the overhead of memory logging becomes insignificant. 

#### A. <100 toks/sec slowdown after introducing the memory probes for Mistral, 

| model_name_or_path | num<br>gpus | per<br>device<br>batch<br>size | throughput<br>with<br>no mem probe<br>(toks/sec) | throughput<br>with<br>mem probe<br>(toks/sec) |
| -------- | -------- | -------- | -------- | -------- |
| mistralai/Mistral-7B-v0.1 | 1 | 4 | 3465 | 3432 |
| mistralai/Mistral-7B-v0.1 | 2 | 2 | 2973 | 2931 |
| mistralai/Mistral-7B-v0.1 | 1 | 8 | 3489 | 3508 |
| mistralai/Mistral-7B-v0.1 | 2 | 4 | 3383 | 3298 |

#### B. <100 toks/sec slowdown after introducing the memory probes for Mixtral
| model_name_or_path | num<br>gpus | per<br>device<br>batch<br>size | throughput<br>with<br>no_mem_probe<br>(toks/sec) | throughput<br>with<br>mem_probe<br>(toks/sec) |
| -------- | -------- | -------- | -------- | -------- |
| mistralai/Mixtral-8x7B-Instruct-v0.1 | 1 | 4 | 1785 | 1776 |
| mistralai/Mixtral-8x7B-Instruct-v0.1 | 2 | 2 | 1518 | 1442 |
| mistralai/Mixtral-8x7B-Instruct-v0.1 | 1 | 8 | 1938 | 1933 |
| mistralai/Mixtral-8x7B-Instruct-v0.1 | 2 | 4 | 1757 | 1724 |

### 2. Torch/HF shows more granularity of memory usage with peak memory and actual allocated memory than Nvidia's reserved memory. This is more helpful when analyzing the actual memory allocated for each model.

We compare the 2 memory tracking methods (Nvidia vs Torch/HF) on single devices for both **GPTQ-LoRA** and **QLoRA**. Nvidia's `peak mem reserved` reports larger values than Torch/HF `peak mem alloc`,  `torch mem alloc` shows the actual memory usage is lesser.

| model_name_or_path | num<br>gpus | per<br>device<br>batch<br>size | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) |
| -------- | -------- | -------- | ------------------------ | --------------------- | --------------------- |
| mistralai/Mistral-7B-v0.1     | 1        | 4                           | 19.46               | 15.86                         | 4.84                     |
| TheBloke/Mistral-7B-v0.1-GPTQ | 1        | 4                           | 19.97               | 15.89                         | 4.87             |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | 1        | 4                           | 37.49               | 36.22                         | 25.2              |
| TheBloke/Mixtral-8x7B-Instruct-v0.1-GPTQ | 1        | 4                           | 36.59               | 35.53                         | 24.51    |
| NousResearch/Llama-2-70b-hf | 1        | 4                           | 71.12               | 68.16                         | 37.35                   |
| TheBloke/Llama-2-70B-GPTQ   | 1        | 4                           | 70.51               | 65.9                          | 36.29                    |

### 3. Memory Usage Decreases on Distributed Finetuning
When running large models on multiple devices, `torch mem alloc` shows the memory usage decreases as the models are sharded (Comparing to table above).
| model_name_or_path | num<br>gpus | per<br>device<br>batch<br>size | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) |
| -------- | -------- | -------- | ------------------------ | --------------------- | --------------------- |
| mistralai/Mistral-7B-v0.1     | 2        | 4                           | 20.97               | 16.59                         | 2.73                     |
| TheBloke/Mistral-7B-v0.1-GPTQ | 2        | 4                           | 23.75               | 16.26                         | 3.01             |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | 2        | 4                           | 32.59               | 29.33                         | 13.22            |
| TheBloke/Mixtral-8x7B-Instruct-v0.1-GPTQ | 2        | 4                           | 51.1                | 27.74                         | 12.79     |

Verified that `torch mem alloc` for GPTQ-LoRA on Llama2-70B hovers at 19GiB when sharded after `trainer.prepare` and during training. The values are similar to the manual probed values from this #15.

| model_name_or_path | num<br>gpus | per<br>device<br>batch<br>size | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) |
| -------- | -------- | -------- | ------------------------ | --------------------- | --------------------- |
| NousResearch/Llama-2-70b-hf | 2        | 2                           | 51.49               | 46.52                         | 19.17                   |
| TheBloke/Llama-2-70B-GPTQ   | 2        | 2                           | 78.69               | 45.4                          | 18.65                    |

## Benchmarks

Run `tox -e run_benches` to produce benchmarks. Full benchmark details can be referenced [here](https://github.com/achew010/fms-acceleration/blob/memory-benchmarks/scripts/benchmarks/refs/a100_80gb.csv)

### 4.  For small models, LoRA runs faster than the Quantized PEFT methods. One likely reason could be that it doesnt require an additional dequantization operation before the base layer+LoRA matmuls. While this is so, we also observe the significantly larger memory it consumed compared to the Quantized PEFT methods.

| model_name_or_path | Training<br>Type | Accel.<br>Config<br>Type | num<br>gpus | per<br>device<br>batch<br>size  | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) | throughput<br>(toks/sec) |
| ----------------------------- | ----------- | ----------- | ----------- | ------------------------- | -------- | --------------------------- | ------------------- | ----------------------------- |
| mistralai/Mistral-7B-v0.1     | lora        | none                      | 1        | 4                           | 29.03               | 26.11                         | 15.12                    | 3597                    |
| mistralai/Mistral-7B-v0.1     | lora        | accelerated-peft-bnb      | 1        | 4                           | 19.46               | 15.86                         | 4.84                     | 3428                    |
| TheBloke/Mistral-7B-v0.1-GPTQ | lora        | accelerated-peft-autogptq | 1        | 4                           | 19.97               | 15.89                         | 4.87                     | 3254                    |

### 5. We observe that on single device finetuning for larger models (e.g. 49B Mixtral), PEFT begins to run out of memory while the Quantized PEFT methods continue to maintain low memory consumption.

| model_name_or_path | Training<br>Type | Accel.<br>Config<br>Type | num<br>gpus | per<br>device<br>batch<br>size  | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) | throughput<br>(toks/sec) |
| ----------------------------- | ----------- | ----------- | ----------- | ------------------------- | -------- | --------------------------- | ------------------- | ----------------------------- |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | none        | none                      | 1        | 4                           | 79.14               | OOM                           | OOM                      | OOM                     |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | lora        | none                      | 1        | 4                           | 79.06               | OOM                           | OOM                      | OOM                       |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | lora        | baseline-peft-bnb         | 1        | 4                           | 47.18               | 46.42                         | 25.73                    | 1396                    |
| mistralai/Mixtral-8x7B-Instruct-v0.1     | lora        | accelerated-peft-bnb      | 1        | 4                           | 37.5                | 36.22                         | 25.2                     | 1764                    |
| TheBloke/Mixtral-8x7B-Instruct-v0.1-GPTQ | lora        | accelerated-peft-autogptq | 1        | 4                           | 36.58               | 35.53                         | 24.51                    | 1864                    |

### 6. In distributed finetuning for large models like Llama2-70B, GPTQ-LoRA shows the lowest memory consumption with the same throughput.

| model_name_or_path | Training<br>Type | Accel.<br>Config<br>Type | num<br>gpus | per<br>device<br>batch<br>size  | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) | throughput<br>(toks/sec) |
| ----------------------------- | ----------- | ----------- | ----------- | ------------------------- | -------- | --------------------------- | ------------------- | ----------------------------- |
| NousResearch/Llama-2-70b-hf | lora        | none                      | 2        | 2                           | 79                  | OOM                             | OOM                        | OOM                       |
| NousResearch/Llama-2-70b-hf | lora        | accelerated-peft-bnb      | 2        | 2                           | 51.4                | 46.52                         | 19.17                    | 418                     |
| TheBloke/Llama-2-70B-GPTQ   | lora        | accelerated-peft-autogptq | 2        | 2                           | 78.5                | 45.4                          | 18.65                    | 426                     |

### 7. Increasing the batch size, GPTQ-LoRA is the only experiment that doesnt run out of memory.
| model_name_or_path | Training<br>Type | Accel.<br>Config<br>Type | num<br>gpus | per<br>device<br>batch<br>size  | peak nvidia mem reserved<br>(GiB) | peak torch mem alloc<br>(GiB) | torch mem alloc<br>(GiB) | throughput<br>(toks/sec) |
| ----------------------------- | ----------- | ----------- | ----------- | ------------------------- | -------- | --------------------------- | ------------------- | ----------------------------- |
| NousResearch/Llama-2-70b-hf | lora        | none                      | 2        | 4                           | OOM                 | OOM                           | OOM                      | OOM                     |
| NousResearch/Llama-2-70b-hf | lora        | accelerated-peft-bnb      | 2        | 4                           | OOM                 | OOM                           | OOM                      | OOM                     |
| TheBloke/Llama-2-70B-GPTQ   | lora        | accelerated-peft-autogptq | 2        | 4                           | 78.48               | 70.67                         | 18.65                    | 451                     |


scripts/benchmarks/benchmark.py
```diff
@@ -308,6 +316,38 @@ def run(self, run_cmd: str, environment_variables: Dict = None):
...
+            '''
+            assert torch.cuda.device_count()>0, "No device detected for memory logging!"
+            nvidia_logging_cmd = [
+                "nvidia-smi",
+                "--query-gpu",
+                GPU_TABLE,
+                "--format",
+                "csv",
+                "--id",
+                str(environment_variables['CUDA_VISIBLE_DEVICES']),
```
fabianlim, line 341: we need to be careful of this
- outside we set `CUDA_VISIBLE_DEVICES` in environment_variables to 0,1...num_gpus
- then here we take them from the environment of `benchmark.py` which can be arbitrary, e.,g. `CUDA_VISIBLE_DEVICES=3,4,6,...`
- so we must be sure that doing this 3 -> 0, 4 -> 1, etc.
- and we must document this down if it is correct

when you print the CSV in the `gpu_log_fileanme`, will be gpu ids be printed as 3,4,6.. or 0,1,2?

also what if CUDA_VISIBLE_DEVICES is not set?

scripts/benchmarks/benchmark.py
```diff
@@ -374,12 +416,29 @@ def maybe_get_experiment_error_traceback(self):
...
+        return mem_usage_by_device_id
+
     def write_result(self):
         "Function to write a json result file"
 
         # save some basic args
         save_result = ConfigUtils.convert_args_to_dict(self.experiment_args_str)
-        save_result["num_gpus"] = self.num_gpus
+        save_result['num_gpus'] = self.num_gpus
+        save_result[REPORT_GPU_FIELD_NAME] = self.get_avg_mem_usage_per_sec_by_device_id().mean()
```
fabianlim, line 441: can we also save the name of the GPU, e.g., A100...

fabianlim (changes requested):
will be good to put a link to this issue soemwhere in the code so that we can view the gpu log schema.

scripts/benchmarks/benchmark.py
```diff
@@ -374,12 +416,29 @@ def maybe_get_experiment_error_traceback(self):
...
+        This function retrieves the gpu memory logs and returns the average memory consumed per device across the experiment
+        Returns a pd.Series of avg mem usage per sec in MiB for each device id
+        ''' 
+        grouped_indices = pd.read_csv(self.gpu_log_filename, skipinitialspace=True).groupby(by='index')
+        # Calculate the average memory consumption per sec in each device
+        mem_usage_by_device_id = grouped_indices.apply(
+            lambda x: x[GPU_LOG_USED_MEM_COLUMN_NAME].str.replace(GPU_LOG_METRIC_SUFFIX, '').astype(float)
+            )
+        # filter only mem measurements that are above a min MiB value before taking the average mem usage per device
+        mem_usage_by_device_id = mem_usage_by_device_id[mem_usage_by_device_id>min_measurement_in_mib]
```
fabianlim, line 430: not very sure why this filtering is required.

achew010: Initial measurement values range from 0-3mb even when the subprocess has started, im filtering out the zero values before i take the mean. If we  use the timestamps difference, it might still have zero values in the series when we take the mean are we okay with that?

fabianlim: @achew010 Ok i see what you are getting at. THen I propose the following

y_i above is not the difference between timestamps, it is the difference between the initial GPU mem usage, and the one at the i-th measurement
e.g., if x_0 = 10mb, and x_i = 40mb, then the difference is 30mb

So then if you have a seuqnece x = 10, 10, 20, 30, 40, 40, .... 30, 20, 10
then x_0 = 10 and so

y = 0, 0, 10, 20, 30, 30, ..., 20, 10, 0

Then filter out another within a small range, say 0-3mb

then y_filtered = 10, 20, 30, 30, ... , 20, 10. 

Then report the final as the mean of y_filtered.

scripts/benchmarks/benchmark.py
```diff
@@ -634,8 +644,21 @@ def _nunique(series):
 def compress(df):
     return df.loc[:, df.apply(pd.Series.nunique) != 1]
 
+def get_available_gpu_ids():
```
fabianlim, line 647: not sure if this warrants a functions ince its pretty simple logic.

scripts/benchmarks/benchmark.py
achew010: Okay i amended the code for all experiments to reference their device ids from `benchmark.py`'s `CUDA_VISIBLE_DEVICES` environment variable and if the env var is not set, default to taking the device ids from an enumeration of `torch.cuda.device_count`

scripts/benchmarks/benchmark.py
achew010: @fabianlim
I edited the function to subtract the first values from the measurements and return the max value for each device id

fabianlim:
can you run `tox -e lint` from the top level directory

scripts/benchmarks/benchmark.py
```diff
@@ -644,24 +659,20 @@ def _nunique(series):
...
-    if no gpu ids are specified, it will default to the enumeration of available ids
-    '''
+def main(args):
+    
+    # Gathers available gpu device ids that will be used for benchmarking.
+    # If "CUDA_VISIBLE_DEVICES" is specified, it will return the specified device ids
+    # if no gpu ids are specified, it will default to the enumeration of available ids
     assert torch.cuda.device_count()>0, "No device detected for memory logging!"
     specified_device_ids = os.environ.get("CUDA_VISIBLE_DEVICES")
     if specified_device_ids:
```
fabianlim, line 669: just rename `specified_device_ids` to `available_gpu_indices`

achew010: okay renamed.

achew010:
> can you run `tox -e lint` from the top level directory, the linting is not automated yet #7 . Also do we plan to activate the memory by default in `run_benchmarks.sh`?



> can you run `tox -e lint` from the top level directory, the linting is not automated yet #7 . Also do we plan to activate the memory by default in `run_benchmarks.sh`?

okay linted. Do we want to set nvidia-smi to be the default mem logging approach or once we establish that the speed degradation is significant with using the HF memory logging API to use [this](https://github.com/achew010/fms-acceleration/tree/hf-memory-benchmarks) instead?

fabianlim:
> okay linted. Do we want to set nvidia-smi to be the default mem logging approach or once we establish that the speed degradation is insignificant with using the HF memory logging API to use [this](https://github.com/achew010/fms-acceleration/tree/hf-memory-benchmarks) instead?
@achew010 lets merge that commit on top of this one and let me review.

scripts/benchmarks/README.md
```diff
@@ -92,3 +92,102 @@ python benchmark.py --help
...
+```
+
+To compute the total GPU memory allocated for training
+- Total memory delta for train stage = 126487040 bytes ('train_mem_gpu_alloc_delta') + 11016635392 bytes ('train_mem_gpu_peaked_delta')
+- Memory consumption for training = 4747206656 bytes ('before_init_mem_gpu') + 0 bytes ('init_mem_gpu_alloc_delta') + 0 bytes ('init_mem_gpu_peaked_delta') + 126487040 bytes ('train_mem_gpu_alloc_delta') + 11016635392 bytes ('train_mem_gpu_peaked_delta')
+
+#### No Significant Slowdown Using HF Memory Probes 
+| acceleration type         | model_name_or_path                       | num_gpus | batch size | throughput without mem probs (toks/s) | throughput with mem probs (toks/s) | allocated gpu memory (GiB) |
+| ------------------------- | ---------------------------------------- | -------- | ---------- | -------------------------- | ---------------------------------- | -------------------------- |
+| accelerated-peft-bnb      | mistralai/Mistral-7B-v0.1                | 1        | 4          | 3385                       | 3451                               | 15.9                       |
```
fabianlim, line 170: summarize for the reader, this is too many numbers and the reader will get confused.

scripts/benchmarks/benchmark.py
```diff
@@ -578,6 +718,19 @@ def compress(df):
...
+    # if no gpu ids are specified, it will default to the enumeration of available ids
+    assert torch.cuda.device_count() > 0, "No device detected for memory logging!"
+    available_gpus_indices = os.environ.get("CUDA_VISIBLE_DEVICES")
+    if available_gpus_indices:
+        available_gpus_indices = available_gpus_indices.split(",")
+    else:
+        available_gpus_indices = [str(i) for i in range(torch.cuda.device_count())]
+
+    if args.dry_run and args.log_nvidia_smi:
+        setattr(args, "log_nvidia_smi", False)
```
fabianlim, line 732: dont use `setattr`, just set it directly. or args.__dict["log_nvidia_smi"]`

scripts/benchmarks/benchmark.py
```diff
@@ -374,13 +461,65 @@ def maybe_get_experiment_error_traceback(self):
...
+        # if a gpu log file exist, process the raw nvidia logs and write to result
+        if os.path.isfile(self.gpu_log_filename):
+            # Add GPU info and measurements into the result saving
+            peak_mem_usage_by_device_id, device_name = self.get_peak_mem_usage_by_device_id()
+            save_result[REPORT_DEVICE_FIELD_NAME] = device_name
+            # Memory usage is averaged across all devices in the final result
+            save_result[REPORT_GPU_FIELD_NAME] = peak_mem_usage_by_device_id.mean()
+
+        # process gpu mem from output metrics and write to result
+        if save_result.get(ARG_SKIP_MEMORY_METRIC) == "False":
```
fabianlim, line 516: call this `HF_ARG_SKIP_MEMORY_METRIC`

scripts/benchmarks/benchmark.py
```diff
@@ -374,13 +461,65 @@ def maybe_get_experiment_error_traceback(self):
...
+            save_result[RESULT_FIELD_ALLOCATED_GPU_MEM] = "%.2f" % sum(
```
fabianlim, line 517: i prefer to use f-strings if possible

scripts/benchmarks/benchmark.py
```diff
@@ -441,6 +580,7 @@ def maybe_get_experiment_error_traceback(self):
 def prepare_arguments(args):
     defaults = ConfigUtils.read_yaml(args.defaults_config_path)
     defaults["training_data_path"] = args.dataset_save_path
+    defaults["skip_memory_metrics"] = not args.log_memory_hf
```
fabianlim, line 583: rather than change the defaults, I think you should handle this the same way `"output_dir"` is handled.

scripts/benchmarks/benchmark.py
```diff
@@ -72,6 +73,49 @@
...
+    NOTE:
+    - HFTrainer only takes the rank0 memory when in distributed mode
+    - HFTrainer uses `torch.cuda.memory_allocated()` and `torch.cuda.max_memory_allocated()` underneath
+    - https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.skip_memory_metrics
+
+    Returns a dictionary of keys and gpu memory values (in GiB) 
+    corresponding to the relevant stage and delta keys for training 
+    """
+    return {
+        key: val/GIB_CONVERSION_FACTOR
```
fabianlim, line 115: i suggest not to do any conversion, just return the raw number

scripts/benchmarks/benchmark.py
```diff
@@ -72,6 +73,49 @@
...
+GPU_LOG_USED_MEM_COLUMN_NAME = "memory.used [MiB]"
+GPU_LOG_METRIC_SUFFIX = " MiB"
+GPU_TABLE = "timestamp,name,index,memory.used"
+REPORT_GPU_FIELD_NAME = "reserved_gpu_mem"
+REPORT_DEVICE_FIELD_NAME = "gpu_device_name"
+
+GIB_CONVERSION_FACTOR = 1e9
+HF_TRAINER_LOG_GPU_MEMORY_STAGES = ['before', 'init', 'train']
+ARG_SKIP_MEMORY_METRIC = "skip_memory_metrics"
+RESULT_FIELD_ALLOCATED_GPU_MEM = "alloc_gpu_memory_in_gib"
```
fabianlim, line 86: just return the raw number

scripts/benchmarks/README.md
```diff
@@ -92,3 +92,102 @@ python benchmark.py --help
...
+    'train_mem_cpu_peaked_delta': 0, 
+    'train_mem_gpu_peaked_delta': 11016635392, 
+    'before_init_mem_cpu': 5831098368, 
+    'before_init_mem_gpu': 4747206656, 
+    'epoch': 0.04
+}
+```
+
+To compute the total GPU memory allocated for training
+- Total memory delta for train stage = 126487040 bytes ('train_mem_gpu_alloc_delta') + 11016635392 bytes ('train_mem_gpu_peaked_delta')
```
fabianlim, line 164: we need a link from HF docs to verify that this is indeed the way to intepret these numbers.

scripts/benchmarks/README.md
fabianlim: i suggest to have two concise sections
1. on the slowdown
2. on the differences between nvidia and HF probes.

scripts/benchmarks/README.md
```diff
@@ -92,3 +92,102 @@ python benchmark.py --help
 
 Note:
 - in `run_benchmarks.sh` we will clear the `RESULT_DIR` if it exists, to avoid contaimination with old results. To protect against overwrite, then always run with `NO_OVERWRITE=true`.
+
+## Logging Memory
+
+There are 2 ways to benchmark memory in `benchmark.py`:
```
fabianlim, line 98: the recommended way to run benchmarks is via `tox` which calls the `benchmark.sh`, most people will not call `benchmark.py` directly
- so if benchmark.sh is hardcoded to a single mode, we need to justify why, based on the study here.
- and these instructions is only for advanced users who will run `benchmark.py` directly.

fabianlim (changes requested):

scripts/benchmarks/benchmark.py
```diff
@@ -503,15 +515,15 @@ def write_result(self):
...
             # Add GPU info and measurements into the result saving
             peak_mem_usage_by_device_id, device_name = self.get_peak_mem_usage_by_device_id()
-            save_result[REPORT_DEVICE_FIELD_NAME] = device_name
+            save_result[RESULT_FIELD_DEVICE_NAME] = device_name
             # Memory usage is averaged across all devices in the final result
-            save_result[REPORT_GPU_FIELD_NAME] = peak_mem_usage_by_device_id.mean()
+            save_result[RESULT_FIELD_RESERVED_GPU_MEM] = peak_mem_usage_by_device_id.mean()
 
         # process gpu mem from output metrics and write to result
         if save_result.get(HF_ARG_SKIP_MEMORY_METRIC) == "False":
```
fabianlim, line 523: this will give problem if `HF_ARG_SKIP_MEMORY_METRIC` is not specified in the arguments.

scripts/benchmarks/benchmark.py
```diff
@@ -103,7 +103,7 @@ def extract_gpu_memory_metrics(output_metrics) -> Tuple[float]:
         HF_TRAINER_LOG_GPU_STAGE_INIT,
         HF_TRAINER_LOG_GPU_STAGE_TRAIN,
     ]
-    alloc_running_sum = output_metrics.get(HF_TRAINER_LOG_GPU_STAGE_BEFORE_INIT)
+    alloc_running_sum = output_metrics[HF_TRAINER_LOG_GPU_STAGE_BEFORE_INIT]
```
fabianlim, line 106: this will throw error if the `--skip_memory_metrics` is not specified

scripts/benchmarks/benchmark.py
```diff
@@ -508,8 +508,9 @@ def write_result(self):
             save_result[RESULT_FIELD_RESERVED_GPU_MEM] = peak_mem_usage_by_device_id.mean()
 
         # process gpu mem from output metrics and write to result
-        _experiment_dict = ConfigUtils.convert_args_to_dict(self.experiment_arg)
-        if _experiment_dict.get(HF_ARG_SKIP_MEMORY_METRIC) == False:
+        argument_idx = self.experiment_arg.index(HF_ARG_SKIP_MEMORY_METRIC)
```
fabianlim, line 511: wont this throw error if `HF_ARG_SKIP_MEMORY_METRIC` is not specified?

fabianlim: oic you forced it in line 642.. not ideal.. but ok..

fabianlim (approved):

fabianlim:
@achew010 i approived. after we update the csv we can merge

scripts/benchmarks/refs/a100_80gb.csv
```diff
@@ -1,49 +1,61 @@
...
+,97451958272.0,512.0,0.09,,none,19,4096.0,0.0,0.0,0.0,2e-4,16,0.0,mistralai/Mixtral-8x7B-Instruct-v0.1,2,69848.0,64347637760.0,lora,4,16,q_proj k_proj v_proj o_proj,47311648768.0,0.9383139928181966,-92302839808.0,92306579456.0,47311648256.0,17035988992.0,280.8919,0.854,0.107,1749.855
+,,,,,none,20,,,,,2e-4,16,0.0,NousResearch/Llama-2-70b-hf,1,80917.0,,lora,4,16,q_proj k_proj v_proj o_proj,,,,,,,,,,
+,,,,,none,21,,,,,2e-4,16,0.0,NousResearch/Llama-2-70b-hf,2,80894.0,,lora,2,16,q_proj k_proj v_proj o_proj,,,,,,,,,,
+,,,,,none,22,,,,,2e-4,16,0.0,NousResearch/Llama-2-70b-hf,1,80917.0,,lora,8,16,q_proj k_proj v_proj o_proj,,,,,,,,,,
+,,,,,none,23,,,,,2e-4,16,0.0,NousResearch/Llama-2-70b-hf,2,80979.0,,lora,4,16,q_proj k_proj v_proj o_proj,,,,,,,,,,
+/data/aaron/experimental/test3/scripts/benchmarks/../../sample-configurations/baseline-peft-bnb-nf4-sample-configuration.yaml,4515069952.0,5241734144.0,0.04,True,baseline-peft-bnb,24,8192.0,0.0,0.0,0.0,2e-4,16,0.0,mistralai/Mistral-7B-v0.1,1,27023.0,22825932800.0,lora,4,16,q_proj k_proj v_proj o_proj,5368221184.0,0.9589527130126954,403791872.0,0.0,126487040.0,17457711616.0,178.8061,0.671,0.168,2748.9
```
fabianlim, line 26: need to understand why the full paths are being propagated in here. Previously we did not have them.

scripts/benchmarks/refs/a100_80gb.csv
```diff
@@ -1,49 +1,61 @@
...
-sample-configurations/accelerated-peft-autogptq-sample-configuration.yaml,,True,accelerated-peft-autogptq,46,2e-4,16,0.0,TheBloke/Nous-Hermes-Llama2-70B-GPTQ,1,,lora,8,16,q_proj k_proj v_proj o_proj,,,,,
-sample-configurations/accelerated-peft-autogptq-sample-configuration.yaml,,True,accelerated-peft-autogptq,47,2e-4,16,0.0,TheBloke/Nous-Hermes-Llama2-70B-GPTQ,2,,lora,4,16,q_proj k_proj v_proj o_proj,,,,,
+acceleration_framework_config_file,before_init_mem_cpu,before_init_mem_gpu,epoch,fp16,framework_config,index,init_mem_cpu_alloc_delta,init_mem_cpu_peaked_delta,init_mem_gpu_alloc_delta,init_mem_gpu_peaked_delta,learning_rate,lora_alpha,lora_dropout,model_name_or_path,num_gpus,nvidia_mem_reserved,peak_torch_mem_alloc_in_bytes,peft_method,per_device_train_batch_size,r,target_modules,torch_mem_alloc_in_bytes,train_loss,train_mem_cpu_alloc_delta,train_mem_cpu_peaked_delta,train_mem_gpu_alloc_delta,train_mem_gpu_peaked_delta,train_runtime,train_samples_per_second,train_steps_per_second,train_tokens_per_second
```
fabianlim, line 1: I think we should only commit the memory summary numbers, and not the stuff like `mem_cpu_peaked_delta`

fabianlim:
We have noted the memory keys should be renamed; to be addressed later in https://github.com/foundation-model-stack/fms-acceleration/issues/19

fabianlim:
@achew010 can we move all the memory computation logic out of `write_result` into `gather_report`. That way the `results.json` only holds the raw data. Then `gather_report` can hold all the logic to preprocess the data for human consumption.

fabianlim:
@achew010 also one more consideration is that memory we should only have `huggingface` mem probes in the `benchmark.csv`. This is because the `script.sh` cannot easily replay the `nvidia-smi` measurements.

Issue #10, opened 2024-05-16
Allow BNB Plugin to be Loaded Without PEFT Wrapping


fabianlim:
This issue regards to the [warnings that for QLoRA PeFT we should pass `peft_config` directly to `SFTTrainer`](https://github.com/huggingface/trl/blob/99f2c94b2200927a1dc156f16e012dca11f865e1/trl/trainer/sft_trainer.py#L204) 
- if `no_peft_model: True` then `model_loader` will simply load the BNB model, and some logic needed to set `requires_agumentation: False`
- peft wrapping will be delegated to `trf.SFTTrainer`
- add one test for this new flag in `tests/test_peft_plugin`
- in generate sample configurations update the `CONFIGURATIONS` and `COMBINATIONS`:
  ```python
    CONFIGURATIONS = {
      KEY_AUTO_GPTQ: "plugins/accelerated-peft/configs/autogptq.yaml",
      KEY_BNB_NF4: (
          "plugins/accelerated-peft/configs/bnb.yaml",
          [("peft.quantization.bitsandbytes.quant_type", "nf4")],
      ),
     KEY_BNB_NF4_BASELINE: (
          "plugins/accelerated-peft/configs/bnb.yaml",
          [
              ("peft.quantization.bitsandbytes.quant_type", "nf4"), 
             ("peft.quantization.bitsandbytes.no_peft_model", True), 
           ],
      ),
  }

  COMBINATIONS = [
      ("accelerated-peft-autogptq", (KEY_AUTO_GPTQ,)),
      ("accelerated-peft-bnb-nf4", (KEY_BNB_NF4,)),
      ("baseline-bnb-nf4", (KEY_BNB_NF4_BASELINE,)),
  ]
  ```
- regenerate the benches, update the CSV and  the README

#### configs/bnb.yaml
add a new flag `no_peft_model`
```yaml

# PEFT-related acceleration
peft:

  # quantization-releated acceleration
  # e.g., kernels for quantized base weights
  quantization: 

    # For loading BitsAndBytes quantized layers
    # to serve as 4bit base-weights for LoRA PEFT-tuning.
    # NOTE: currently AutoGPTQ is not properly integrated into huggingface /
    # bitsandbytes, thus recommended quant_type to be either "nf4"
    # or "fp4".
    # bitsandbytes:
    bitsandbytes:
      quant_type: nf4 

      # If True, then no get_peft_model and prepare_model_for_kbit_training
      # will be called. 
[...]

Issue #9, opened 2024-05-16
Finish Linting of Accelerated Peft Plugin and Scripts


fabianlim:
In continuation of #7 , finish and activate the linting for `accelerated_peft` and `scripts/benchmarks/bencmarks.py`

fabianlim:
What remains is to activate the linting for benchmark in the github workflow.

PR #16, opened 2024-05-21
Added support for running official HF baseline FSDP-QLoRA benchmark

achew010:
This PR addresses issue #10 by adding support for a FSDP-compatible HF QLoRA baseline to the our benchmarks. 

### Feature
This will allow users to specify a `no_peft_model` field in the plugin config `bnb.yaml`. Specifying this field will bypass the `plugin.augmentation` function and allow SFTTrainer to manage the PEFT preparation of the model instead. 

#### NOTE:
- While the open-source approach to FSDP-compatible QLoRA removes the extraneous dtype casting in `prepare_model_for_kbit_training`, it only does so when the model is sharded. When on single device, it continues to use `prepare_model_for_kbit_training` and users will continue to experience a slowdown due to the extraneous casting.

### DIFF
--- a/plugins/accelerated-peft/tests/test_peft_plugins.py
+++ b/plugins/accelerated-peft/tests/test_peft_plugins.py
@@ -122,6 +122,20 @@ def test_configure_bnb_plugin():
             assert framework.requires_agumentation
             assert len(framework.get_callbacks_and_ready_for_train()) == 0
 
+    # test no_peft_model is true skips plugin.augmentation
+    for key, correct_value in [
+        ("peft.quantization.bitsandbytes.no_peft_model", True),
+        ("peft.quantization.bitsandbytes.no_peft_model", False),
+    ]:
+        with instantiate_framework(
+            update_configuration_contents(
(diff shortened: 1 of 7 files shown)


plugins/accelerated-peft/tests/test_peft_plugins.py
```diff
@@ -122,6 +122,20 @@ def test_configure_bnb_plugin():
...
+        ("peft.quantization.bitsandbytes.no_peft_model", False),
+    ]:
+        with instantiate_framework(
+            update_configuration_contents(
+                read_configuration(CONFIG_PATH_BNB), key, correct_value
+            ),
+            require_packages_check=False,
+        ):
+            # check flags and callbacks
+            assert (not correct_value)==framework.requires_agumentation
```
fabianlim, line 137: i can see some liniting issues, but we can take care of it in #9

fabianlim (approved):
LGTM save for that linting issue. but im ok to merge this first and then handle #7 later.

PR #29, opened 2024-05-30
Add MLP & QLoRA Fused Ops and Kernels, Mixtral

fabianlim:
Completing more items in #25 .
- decided to remove the L40 benchmarks. 

*Verified that we can reproduce the roughly 20% speedups using fused-ops and kernels* 
- these are per device throughputs, so for two gpus we should multiply by 2 to get the actual througput
image

*Verified that we are reproduce the 75% in memory reduction using 4bit base weights*
- also with FSDP when using two gpus, we see another 50% memory reduction
image


achew010 (approved):

fabianlim:
running a set of benches now. will merge after complete

PR #34, opened 2024-06-07
Updated benchmark reference

achew010:
Updated a100 benchmark numbers after the FSDP-ignore -modules fix


fabianlim:
@achew010 this CSV would correspond to numbers in #31 correct?

achew010:
> @achew010 this CSV would correspond to numbers in #31 correct?

Yes the tables in the description of #31 are referenced from this csv

Issue #32, opened 2024-06-06
Allow Fused Ops to Support Dropout


fabianlim:
Right now the fused ops do not support dropout, but perhaps it can be quite trivally supported as this is the implementation of the dropout in `QuantLinear` in both `peft.tuners.lora.bnb` and `peft.tuners.lora.gptq`
```
output = lora_B(lora_A(dropout(x)))
```

PR #37, opened 2024-06-15
Fused Ops Support for Lora Dropout

fabianlim:
This PR addresses #32 to allow Fused Ops to also support lora dropout.

The key strategy is to pass the `dropout` to the `LoRA.apply` function, and have it call inside the `matmul_lora`, and return out the dropped out `X` to the top-level to be saved for the `backward`
- `LoRA.apply(.., dropout)` : `dropout` is passed into `apply`.
- `matmul_lora(..., dropout)` is called with the `dropout`, in which inside
    ```python
    X = dropout(X)
    dropout.X = X # saved for retuning to the calling level
   ```
- then `ctx.save_for_backward(..., dropout.X)` is saved for backward (and we will delete it off `dropout` when it is no more needed)
- then inside `backward`, we will use `dropout.X` instead of X.


fabianlim:
merged to the wrong branch, so I cherry picked to `dev` on commit 6186ddd4e2dd4c8f53b803bcf032b9fb481ec51e

PR #42, opened 2024-06-26
Bound Package Versions

achew010:
Bounding package versions to avoid recent incompatibilities for
- AutoGPTQ (Building from latest commit require HPC libraries as dependencies)
- numpy (recently updated to `2.0`, this isnt compatible for torch versions `<2.3`)


fabianlim (approved):
LGTM

PR #51, opened 2024-07-15
Publish Accelerated Peft and Fused-Ops Plugins

achew010:
This PR prepares the accelerated-peft and fused-ops-and-kernels plugins for publishing.

Items:
- Change Framework CLI tool to support installation from PyPi
- Set version numbers for both plugin package
- Enable tox build sections for both plugins


fabianlim (approved):
LGTM

PR #64, opened 2024-08-07
Rename Plugin to `AttentionAndDistributedPacking `

achew010:
## Description

This PR renames the InstructLab plugin to AttentionAndDistributedPacking. Functionally, it is the same behaviour as the current plugin but all references of instructlab will be replaced. 

```
export DATASET=benchmark_outputs/data/cache_mistralai_Mistral_7B_v0.1.json
python -m tuning.sft_trainer --model_name_or_path mistralai/Mistral-7B-v0.1 --padding_free huggingface --packing False --max_seq_len 4096 --learning_rate 2e-5 --torch_dtype float16 --use_flash_attn True --include_tokens_per_second True --num_train_epochs 1 --gradient_accumulation_steps 1 --gradient_checkpointing True --evaluation_strategy no --save_strategy no --weight_decay 0.01 --warmup_steps 10 --adam_epsilon 1e-4 --lr_scheduler_type linear --logging_strategy steps --logging_steps 10 --max_steps 100 --training_data_path $DATASET --per_device_train_batch_size 4 --output_dir benchmark_outputs/aadp --skip_memory_metrics False
```


.github/workflows/build-and-publish.yml
```diff
@@ -14,6 +14,7 @@ jobs:
           - "framework"
           - "accelerated-peft"
           - "fused-ops-and-kernels"
```
fabianlim, line 16: @achew010 lets remove this line until we are ready to publish

fabianlim (changes requested):

fabianlim:
we have renamed the `ilab` to `aadp`, we will be deleting the old pypi project https://pypi.org/manage/project/fms-acceleration-ilab/settings/

PR #74, opened 2024-08-23
Fix formatter

achew010:
## Description

This PR formats the code and fixes the format workflow to fail if additional formatting is required in GH Actions


scripts/benchmarks/benchmark.py
```diff
@@ -430,8 +428,8 @@ def is_completed(self):
         # return complete only if no errors
         # and is not a dry run
```
fabianlim, line 429: lets not format anything in `scripts/benchmarks` since there is another PR #69 making changes to these files.

Issue #70, opened 2024-08-19
Inconsistency in Benchmarks with Different Transformers Versions


achew010:
## Description

We observe no improvement with PaddingFree on QLoRA and GPTQ-LoRA when running benchmarks on OrcaMath. 

However, 
- additionally applying FOAK along with PaddingFree shows significant improvement.
- the benchmarks using FLAN also shows an improvement when PaddingFree is used with QLoRA and GPTQ-LoRA.

Mistral7B (OrcaMath)
| Framework Type | Num Device | Device Batch Size | Train Runtime (sec) | Throughput (toks/sec) | % Runtime Improvement |
| -------- | ------- | ------- | ------- | ------- | ------- |
| Full FT | 2 | 4 | 231 | 1188 | base |
| Full FT + PF | 2 | 4 | 177 | 1584 | +23 |
| BNB | 2 | 4 | 392 | 709 |  base |
| BNB + PF | 2 | 4 | 398 | 704 | +0 |
| BNB + FOAK | 2 | 4 | 190 | 1468 | base |
| BNB + FOAK + PF | 2 | 4 | 155 | 1802 | +18 |
| GPTQ | 2 | 4 | 388 | 720 | base |
| GPTQ + PF | 2 | 4 | 386 | 725 | +0 |
| GPTQ + FOAK | 2 | 4 | 186 | 1359 | base |
| GPTQ + FOAK + PF | 2 | 4 | 158 | 1771 | +15 |

achew010:
## Description
We are again observing some variation with paddingfree in benchmarks for #78 comparing against [published benches](https://github.com/foundation-model-stack/fms-acceleration/blob/main/scripts/benchmarks/refs_orca/a100_80gb_pf.csv), there are some instances the improvement on the training speed is significantly lesser.

### FLAN (6000 samples)
For FLAN, we notice current runs for both pretokenized and untokenized datasets show lesser improvement than before.

Experiment | framework config | num device | batch size | ga | runtime | throughput | improvement |
-- | -- | -- | -- | -- | -- | -- | -- 
Current Published | GPTQ-LoRA + FOAK | 2 | 4 | 2 | 1034 | 1372 | baseline
Current Published | GPTQ-LoRA + FOAK + PF | 2 | 4 | 2 | 587 | 2472 | 43%
pretokenized | GPTQ-LoRA + FOAK | 2 | 4 | 2 | 1029 | 1378 | baseline
[...]

PR #78, opened 2024-08-30
Allow PaddingFree to work with DataCollatorForCompletionOnlyLM

fabianlim:
## Description
Currently padding free plugin only works with `DataCollatorWithFlattening`. This PR makes it also work with `DataCollatorForCompletionOnlyLM` that is used for the *non-pretokenized* use case
- in the `scenarios.yaml`, we should be above to use the `chat_templates`, but set `tokenize=False` and remove the `null` settings on the required fields for tokenization.
- verified that full-FT + paddingfree improvement is consistent at 22-23% for [orca benches](https://github.com/foundation-model-stack/fms-acceleration/blob/non-pretok-pf/scripts/benchmarks/refs_orca/a100_80gb_pf.csv)

Note:
- There are some inconsistent values in benchmarks for PF with FOAK tracked in #70.

## Tests on Flan Subset (6000 samples)

1. Verified that dataset `data_cache.json` is only formatted and untokenized. To ensure that the loss is masked, added a keyword '`RESPONSE:`' in the chat template and as the response template that `DataCollatorForCompletionOnlyLM` will use to mask the loss.

     Example extracted from `dataset['train']['output'][0]`
      ```
      'Write the response. A 2 person conversation: -- Who was selected with the 5th pick in the 1974 NBA draft?. --  RESPONSE: Five other players from this draft, 6th pick Scott Wedman, 8th pick Campy Russell , 12th pick Brian Winters, 21st pick Billy Knight and 25th pick John Drew, were also selected to at least one All-Star Game.'
      ```

2. Verified that using an untokenized dataset to SFTTrainer matches previous padding-free performance with a pretokenized dataset.

    **Untokenized FLAN Dataset**
    | Framework Config | Num Devices | Per Device Batch Size | Train Runtime (secs) | Speedups |
    | -------- | ------- | ------- | ------- | ------- |
    | full-FT | 2 | 4 | 1516 | baseline |
    | padding-free | 2 | 4 | 848 | 1.78x |
    | padding-free + multipack | 2 | 4 | 747 | 2.02x |
    
    **Tokenized FLAN Dataset**
    | Framework Config | Num Devices | Per Device Batch Size | Train Runtime (secs) | Speedups |
    | -------- | ------- | ------- | ------- | ------- |
    | full-FT | 2 | 4 | 1537 | baseline |
    | padding-free | 2 | 4 | 859 | 1.79 x |
    | padding-free + multipack | 2 | 4 | 751 | 2.05 x |


scripts/benchmarks/benchmark.py
```diff
@@ -206,9 +206,10 @@ def prepare_dataset(
                 )
             response_template = self.response_template
 
-        if self.kwargs["tokenize"]:
-            tokenizer = AutoTokenizer.from_pretrained(model_name)
+        # pass in tokenizer to use apply_chat_templates
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
 
+        if self.kwargs["tokenize"]:
```
fabianlim, line 212: you enter here on this condition

```
if self.kwargs['tokenize'] or (not self.kwargs['tokenize'] and self.kwargs['chat_template'])
```

scripts/benchmarks/data_processing.py
```diff
@@ -64,6 +64,7 @@ def _build_data_formatting_func(
             "response_field must be specified if tokenize=True and response_template=None."
 
     def _format(example):
+        nonlocal loss_masking # reference to variable in _build_data_formatting_func
```
fabianlim, line 67: put a comment to say that format is going to be passed into `map` , wherby if `loss_masking` is not rebound using `nonlocal`, then the spawned processes will have no reference to it.

fabianlim:
@achew010 can you do a sanity check and open the `data_cache.json` to ensure it was not tokenized when `tokenize=False`

scripts/benchmarks/refs_orca/requirements.txt
```diff
@@ -1,64 +1,40 @@
...
--e git+https://github.com/foundation-model-stack/fms-acceleration.git@8bc65cd3e0f83a7786e4716216fbe1bea702313b#egg=fms_acceleration_aadp&subdirectory=plugins/attention-and-distributed-packing
--e git+https://github.com/foundation-model-stack/fms-acceleration.git@8bc65cd3e0f83a7786e4716216fbe1bea702313b#egg=fms_acceleration_foak&subdirectory=plugins/fused-ops-and-kernels
--e git+https://github.com/foundation-model-stack/fms-acceleration.git@8bc65cd3e0f83a7786e4716216fbe1bea702313b#egg=fms_acceleration_peft&subdirectory=plugins/accelerated-peft
--e git+https://github.com/foundation-model-stack/fms-hf-tuning.git@a8ab68ffaa0d3b49aeb6753bccfdf807672eba69#egg=fms_hf_tuning
+-e git+https://github.com/foundation-model-stack/fms-acceleration.git@4831b35b3f2d8c8116e55faed3d050a51ef962a0#egg=fms_acceleration&subdirectory=plugins/framework
+-e git+https://github.com/foundation-model-stack/fms-acceleration.git@4831b35b3f2d8c8116e55faed3d050a51ef962a0#egg=fms_acceleration_foak&subdirectory=plugins/fused-ops-and-kernels
+-e git+https://github.com/foundation-model-stack/fms-acceleration.git@4831b35b3f2d8c8116e55faed3d050a51ef962a0#egg=fms_acceleration_peft&subdirectory=plugins/accelerated-peft
```
fabianlim, line 21: maybe you need to fix the cli before updating the references, because otherwise the commit will be missing

PR #80, opened 2024-09-02
Create fused_CEloss.py

wynterl:
Create fused CE Loss from Liger kernels available at https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py


fabianlim:
superceeded by #93

Issue #83, opened 2024-09-11
Distributed Training Problems for QLoRA models with Transformers pre-release 4.45 


achew010:
### Root Cause

The root cause is due to recent transformers update [to resolve high CPU usage for large quantized models](https://github.com/huggingface/transformers/pull/33154).
- what the PR does is to when loading the state dict, puts the quantized model on the `meta` device on ranks > 0, and then shards the weights during the preperation step
- however, this will require `torch.distributed` memory calls during `Accelerator.prepare_model`, which we observe it being stuck for QLoRA `Mistral 7B`.
- This is most probably an NCCL issue. THe PR also says that similar issues were encountered on their test platform (AWS), but overcome by upgrading to most recent NCCL. unfortunately, this fix didnt work on our infrastructure. 

### What was observed

Running experiments to test new Granite models (e.g.  `ibm/PowerLM-3b`) available on Transformers==`4.45.0.dev0`. Encountered the following issues;

1. Hanging inside `trainer.train()` leading to an eventual distributed timeout error for FSDP-QLoRA experiments despite only using standard HF libraries in our baseline experiments.
```
[rank0]:[E906 21:06:59.080356778 ProcessGroupNCCL.cpp:1375] [PG 0 (default_pg) Rank 0] First PG on this rank that detected no heartbeat of its watchdog.
[rank0]:[E906 21:06:59.080547547 ProcessGroupNCCL.cpp:1413] [PG 0 (default_pg) Rank 0] Heartbeat monitor timed out! Process will be terminated after dumping debug info. workMetaList_.size()=8
[...]

PR #89, opened 2024-10-08
Fixes to Accelerated Peft

fabianlim:
This PR provides two fixes
1. fallback to target module defaults. To keep in line with regular HF behavior, we fallback to HF default target modules if `--target_modules` happen to be omitted from the `peft_config`
2. disabling the gptq v1 to v2 conversion: we revert back to default HF behavior and disable offloading. Upon reviewing the [accelerate.load_checkpoint_in_model](https://github.com/huggingface/accelerate/blob/127818fc27ebe5cb236357fff59ff1748326d643/src/accelerate/utils/modeling.py#L1611) i know start to feel the offloading may not be needed moving forward. The main purpose of that function is to read the (sharded) checkpoint and load it into the model (on the CPU, since low_cpu_mem_mode is not working for GPTQ). I think the proper way forward is to get `low_cpu_mem_mode` working for GPTQ, and not worry about offloading the state dict (as it is a hack in itself).

However do note that for `low_cpu_mem_mode` we have observed some problems in the BNB case for certain models (e.g., `GraniteCausalForLM`), see https://github.com/foundation-model-stack/fms-acceleration/issues/83

### DIFF
--- a/plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
+++ b/plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
@@ -35,11 +35,28 @@ PATCH_FOR_FSDP_TRITON_V2 = ["qweight", "qzeros"]
 PEFT_ALL_LINEAR = "all-linear"
 
 
-def requires_installation_on_all_linears(peft_config):
+def requires_installation_on_all_linears(peft_config, model_type: str = None):
     tm = peft_config.target_modules
+
+    if tm is None:
+        try:
+            # Third Party
+            # pylint: disable=import-outside-toplevel
+            from peft.utils.constants import (
+                TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING,
+            )
+
+            tm = TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[model_type]
+        except (ImportError, IndexError) as e:
+            raise ValueError(
+                "target modules not specified and unable to determine default "
+                f"for given model type {model_type}."
+            ) from e
+
+        # replace with some defaults
     assert isinstance(
         tm, (list, set, str)
-    ), "target modules can only be list, set or string"
+    ), "if provided, target modules can only be list, set or string"
     if isinstance(tm, (list, set)):
         if PEFT_ALL_LINEAR not in tm:
             return False
--- a/plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
+++ b/plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
@@ -47,7 +47,7 @@ def requires_installation_on_all_linears(peft_config, model_type: str = None):
             )
 
             tm = TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[model_type]
-        except (ImportError, IndexError) as e:
+        except (ImportError, KeyError) as e:
             raise ValueError(
                 "target modules not specified and unable to determine default "
                 f"for given model type {model_type}."
--- a/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py
+++ b/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py
@@ -319,10 +319,18 @@ class AutoGPTQAccelerationPlugin(AccelerationPlugin):
             )
 
         # Install GPTQ adapters using the AutoGPTQ package (with the above patches)
+        # - try to get a model config to auto determine the layers that can be helpful
+        #
+        model_type = (
+            model.config.model_type if hasattr(model.config, "model_type") else None
+        )
         model = get_gptq_peft_model(
             model,
             peft_config=peft_config,
-            auto_find_all_linears=requires_installation_on_all_linears(peft_config),
+            auto_find_all_linears=requires_installation_on_all_linears(
+                peft_config,
+                model_type=model_type,
+            ),
             train_mode=True,  # install adapaters for training
         )
 
--- a/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_bnb.py
+++ b/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_bnb.py
@@ -144,8 +144,7 @@ class BNBAccelerationPlugin(AccelerationPlugin):
             # that low_cpu_mem_mode=True can cause torch distributed primitives
             # to hang.
 
-            # if _transformers_version >= "4.45":
-            if False:
+            if _transformers_version >= "4.45":
 
                 # pylint: disable=import-outside-toplevel
                 # Third Party
--- a/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_bnb.py
+++ b/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_bnb.py
@@ -144,7 +144,8 @@ class BNBAccelerationPlugin(AccelerationPlugin):
             # that low_cpu_mem_mode=True can cause torch distributed primitives
             # to hang.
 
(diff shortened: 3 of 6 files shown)


plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
```diff
@@ -35,11 +35,28 @@
...
+
+    if tm is None:
+        try:
+            # Third Party
+            # pylint: disable=import-outside-toplevel
+            from peft.utils.constants import (
+                TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING,
+            )
+
+            tm = TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[model_type]
```
willmj, line 49: Hi Fabian, thanks for the quick fix! I was curious - why do we want to set this here when [similar code](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/model.py#L467) is called in the PEFT library already?

fabianlim: this is because for GPTQ, we do not use the `peft` version of `LoraModel`, we use the `GPTQLoRAModel` https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/utils/peft.py#L46

This. is due to the different handling of the quantized base layer

willmj: Thanks for the clarification, that's good to know! In that case, this looks good to me but I'd like to get another set of eyes on it before merging.

plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
```diff
@@ -35,11 +35,28 @@
...
+        except (ImportError, IndexError) as e:
```
anhuong, line 50: If the model_type isn't able to be found then it will pass it in as None and then when trying to do `TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[None]` this leads to a KeyError. I think this should also be captured in this error. The same would happen if a model_type is given that is not in the dictionary. I don't know if we would hit IndexError

anhuong (commented):
The changes for removing offload_state_dict and offload_buffers looks good as well as using the mapping from Peft. Small comment on the error and I can try to test this change on a small model to ensure it works with setting target_modules=None. Also note there are a few small changes to the benchmark scenarios for power granite models

anhuong:
I confirmed on my pod with fms-acceleration and fms-acceleration-peft plugins that I got error `AssertionError: target modules can only be list, set or string`.

I installed fms-acceleration both the framework plugin and the accelerated-peft plugin from this branch. I then ran qlora with config below and it ran successfully:
```json
{
  "model_name_or_path": "/fmaas-integration-tests/models/granite-8b-code-instruct-gptq/",
  "training_data_path": "/fmaas-integration-tests/tuning/input/twitter-complaints.json",
  "output_dir": "/fmaas-integration-tests/tuning/output/anhuong/granite-20b-gptq-qlora-twitter_202401008_1706",
  "num_train_epochs": 10.0,
  "per_device_train_batch_size": 4,
  "gradient_accumulation_steps": 1,
  "learning_rate": 1e-4,
  "response_template": "\n### Label:",
  "dataset_text_field": "output",
  "peft_method": "lora",
  "r": 8,
  "lora_dropout": 0.05,
  "lora_alpha": 16,
  "max_seq_length": 4096,
  "lora_post_process_for_vllm": true,
  "auto_gptq": ["triton_v2"],
  "torch_dtype": "float16",
  "fp16": true
}
```

and it ran successfully. In the adapter_config.json you can see it used the default target modules:
```json
"target_modules": [
    "v_proj",
    "q_proj"
  ],
```

I do see these messages and I'm curious what `autogptq_patch_tensors_as_float_parameters` is?
```sh
***** FMS AccelerationFramework *****
INFO:framework.py:***** FMS AccelerationFramework *****
Active Plugin: AutoGPTQAccelerationPlugin. Python package: fms_acceleration_peft. Version: 0.3.0.dev0.
INFO:framework.py:Active Plugin: AutoGPTQAccelerationPlugin. Python package: fms_acceleration_peft. Version: 0.3.0.dev0.
INFO:framework.py:***************** Module Forwards Patching *************
Rule: autogptq_patch_tensors_as_float_parameters Module: base_layer                Class: QuantLinear     Num: 72
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: base_layer                Class: QuantLinear     Num: 72
Rule: autogptq_patch_tensors_as_float_parameters Module: down_proj                 Class: QuantLinear     Num: 36
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: down_proj                 Class: QuantLinear     Num: 36
Rule: autogptq_patch_tensors_as_float_parameters Module: gate_proj                 Class: QuantLinear     Num: 36
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: gate_proj                 Class: QuantLinear     Num: 36
Rule: autogptq_patch_tensors_as_float_parameters Module: k_proj                    Class: QuantLinear     Num: 36
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: k_proj                    Class: QuantLinear     Num: 36
Rule: autogptq_patch_tensors_as_float_parameters Module: o_proj                    Class: QuantLinear     Num: 36
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: o_proj                    Class: QuantLinear     Num: 36
Rule: autogptq_patch_tensors_as_float_parameters Module: up_proj                   Class: QuantLinear     Num: 36
INFO:framework.py:Rule: autogptq_patch_tensors_as_float_parameters Module: up_proj                   Class: QuantLinear     Num: 36
INFO:framework.py:***************** Accelerator Patching *************
Currently training with a batch size of: 4
***** Running training *****
  Num examples = 50
  Num Epochs = 10
  Instantaneous batch size per device = 4
  Total train batch size (w. parallel, distributed & accumulation) = 8
  Gradient Accumulation steps = 1
  Total optimization steps = 70
  Number of trainable parameters = 1,916,928
  0%|                                                                              | 0/70 [00:00<?, ?it/s]Detected flash_attn version: 2.6.3
{'loss': 6.4873, 'grad_norm': 3.5396203994750977, 'learning_rate': 9e-05, 'epoch': 1.0}       
```

plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py
fabianlim: Ah thanks for his catch. i intended to use `KeyError` but somehow mistook it as an `IndexError`. The latter is for arrays but here we are dealing with a dict

fabianlim:
@anhuong `autogptq_patch_tensors_as_float_parameters` is a [rule](https://github.com/foundation-model-stack/fms-acceleration/blob/ff14d4f0c4df1d7b89b1ad573bcbd235194d4fc4/plugins/accelerated-peft/src/fms_acceleration_peft/autogptq_utils.py#L71) [registered](https://github.com/foundation-model-stack/fms-acceleration/blob/ff14d4f0c4df1d7b89b1ad573bcbd235194d4fc4/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py#L217) only when the world_size > 2. This is for FSDP sharding of the quantized weights

Quant weights are in `int32`, whereby FSDP only supports float types (I think its because `torch.distributed` primitives do not support non-floats). Hence 
- we need to upfront convert them to float types (in this case `torch_dtype`, which is the type passed by user).
- then before the forward call, we need to `view` them into the `int32` dtype, so it can be passed to the dequant.

anhuong (approved):
Thanks Fabian for the changes and the explanation! This looks great!

Issue #76, opened 2024-08-29
Introduce Liger Fused Cross Entropy Kernel to FOAK Plugin 


achew010:
## Description

Consider adding additional Fused CrossEntropyLoss Kernel to FOAK set of kernels

A comparison of the current [FOAK kernels](https://github.com/foundation-model-stack/fms-acceleration/tree/main/plugins/fused-ops-and-kernels) against the kernels from [Liger](https://github.com/linkedin/Liger-Kernel/tree/main) using Liger's [benchmark script](https://github.com/achew010/Liger-Kernel/tree/foak-benchmarks) with the following parameters;

- model: "meta-llama/Meta-Llama-3-8B"
- dataset: 10000 sample subset of "tatsu-lab/alpaca"
- max_seq_length: 512

![image](https://github.com/user-attachments/assets/c6145cd2-826e-4500-8f0c-37e1662ff8b5)

fabianlim:
@achew010 currently we only install the non-chuncked CE loss from unsloth. It seems to be OK for llama3, but scaling up, we should consider handling chunking also

anhuong:
Implemented (2) in this PR https://github.com/foundation-model-stack/fms-acceleration/pull/93 but as noted above it suffers from maintainability as
1. llama and mistral have very similar forward functions but mixtral has different forward function that also needs to be copied in and granite has a slight difference
2. hard to maintain between different transformers versions
3. was not able to show train_runtime improvement from other triton kernels but was able to show memory improvement (as noted in the PR)

Worked on implementing (3) via PR https://github.com/foundation-model-stack/fms-acceleration/compare/main...anhuong:fms-acceleration:fused-cross-entropyloss?expand=1 however the model patching rules aren't quite right here.

Thus it could be worth it to implement (2) via the PR for just transformers v4.44/4.43 and then implement the new solution with the model patching changes needed form #98.

PR #93, opened 2024-10-16
feat: add liger kernel with fused cross entropy loss

anhuong:


plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/liger/fused_linear_cross_entropy_loss.py
```diff
@@ -0,0 +1,577 @@
...
+            target,
+            bias,
+            self.ignore_index,
+            self.label_smoothing,
+            self.reduction,
+        )
+
+# TODO: how to add diff docstrings for diff model types? what if the loss functions aren't the same across models?
+# @add_start_docstrings_to_model_forward(LLAMA_INPUTS_DOCSTRING)
+@replace_return_docstrings(
```
anhuong, line 302: Question: do we need this`add_start_docstrings_to_model_forward` header that exists in https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L1133 but is different for each model type?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/liger/fused_linear_cross_entropy_loss.py
```diff
@@ -0,0 +1,577 @@
...
+            logits = [
+                F.linear(hidden_states, lm_head_slices[i])
+                for i in range(self.config.pretraining_tp)
+            ]
+            logits = torch.cat(logits, dim=-1)
+        else:
+            # TODO: differing line below in granite models compared to llama/mistral model type
+            # logits = logits / self.config.logits_scaling
+            # Only compute necessary logits, and do not upcast them to float if we are not computing the loss
+            logits = self.lm_head(hidden_states[:, -num_logits_to_keep:, :])
```
anhuong, line 412: The forward methods are very similar for llama, granite, and mistral however the above line as you can see is within granite here: https://github.com/huggingface/transformers/blob/v4.45.2/src/transformers/models/granite/modeling_granite.py#L1102 --> how to handle this difference?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/liger/fused_linear_cross_entropy_loss.py
```diff
@@ -0,0 +1,577 @@
...
+        past_key_values=outputs.past_key_values,
+        hidden_states=outputs.hidden_states,
+        attentions=outputs.attentions,
+    )
+
+# TODO: is adding a separate copy of lce_forward() the right path or should the additional logic for Moe models be in the single lce_forward?
+@add_start_docstrings_to_model_forward(MIXTRAL_INPUTS_DOCSTRING)
+@replace_return_docstrings(output_type=MoeCausalLMOutputWithPast, config_class=_CONFIG_FOR_DOC)
+# Ignore copy
+def lce_forward_mixtral(
```
anhuong, line 444: Added a different forward method for mixtral since it was different than the others in line 550-561 and in the object returned is of different type. Is adding a separate forward method the best way to handle this?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/models/granite.py
```diff
@@ -40,6 +41,7 @@ def get_mp_rules(base_type: str):
     try:
         # Third Party
         from transformers.models.granite.modeling_granite import (  # pylint: disable=import-outside-toplevel
+            GraniteForCausalLM,
```
anhuong, line 44: I'm guessing granite here refers to teh GraniteForCausalLM models and the granite with llama architecture models would be handled within the llama.py correct? I'm not sure if the liger kernels work for any of the granite models, do you know?

anhuong:
Will update with benchmark data as I gather them, currently I do see the new kernel being used:
```sh
***** FMS AccelerationFramework *****
INFO:framework.py:***** FMS AccelerationFramework *****
Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.2.1.dev0.
INFO:framework.py:Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.2.1.dev0.
INFO:framework.py:***************** Module Forwards Patching *************
Rule: llama-fused-lce Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-fused-lce Module:                           Class: LlamaForCausalLM Num:  1
Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
INFO:framework.py:Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:***************** Accelerator Patching *************
Currently training with a batch size of: 2
max_steps is given, it will override any value given in num_train_epochs
/home/tuning/.local/lib/python3.11/site-packages/trl/trainer/sft_trainer.py:421: UserWarning: You passed `packing=True` to the SFTTrainer/SFTConfig, and you are training your model with `max_steps` strategy. The dataset will be iterated until the `max_steps` are reached.
  warnings.warn(
***** Running training *****
  Num examples = 2,364
  Num Epochs = 1
  Instantaneous batch size per device = 2
  Total train batch size (w. parallel, distributed & accumulation) = 4
  Gradient Accumulation steps = 1
  Total optimization steps = 100
  Number of trainable parameters = 1,703,936
```

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/models/granite.py
fabianlim: thats the problem with this solution, because we copy and paste the forward functions wholesale. so if the forwards are different this doesnt work

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/liger/fused_linear_cross_entropy_loss.py
fabianlim: the naive method is to copy and modify multiple forward functions for the fused kernel. The better solution is the other design

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/liger/fused_linear_cross_entropy_loss.py
fabianlim: I thinik dont need

sample-configurations/foak-fast-kernels-sample-configuration.yaml
```diff
@@ -22,10 +22,13 @@ plugins:
...
+      fast_loss: False
 
       # fast rms norm triton kernels
       fast_rsm_layernorm: True
 
       # fast RoPE embedding triton kernels
       fast_rope_embeddings: True
+
+      # fused linear cross entropy loss
+      fused_linear_loss: True
```
fabianlim, line 34: I think this is problematic because `fast_loss` and `fused_linear_loss` both switch the loss, so it causes weird combinations

Maybe its better to convert `fast_loss` to accept both boolean and a string

`fast_loss: True` means legacy `fast_loss: 'fused'` means fused

fabianlim: or use `fast_loss` = True / False and `fast_loss_type` we can have the deafult (non-fused) and `"fused"`

anhuong: But that's why we are only allowing the user to set one, if one is true, the other must be false. That prevents this issue right?

fabianlim: I just feel that if already having `True/False` switches is already abit annoying as you mentioned before. Now having constraints between them might be even more confusing

fabianlim:
@anhuong it is a bit unexpected that you are not seeing speedups, as it was reported previously for the same model (alibet different seq lengths are batch sizes) that there are significant speedups https://github.com/foundation-model-stack/fms-acceleration/issues/76

Even though the bench parameters that you tried are different, in ours we actually pack the sequences to full 2048, so we expect that this packing will end up with similar throughputs

Once difference I can see are that yours are multi gpu runs, can we first look at the issue using 1 gpu first.

fabianlim:
@anhuong i was able to repro the liger benches by running them like this

```
cd $HOME/Liger-Kernel/examples/huggingface
pip install -r requirements.txt
bash run_benchmarks.sh
```

and doing some hacking to enable / disable the fused_cross entropy loss

image
image

anhuong:
For reference, using the fms-acceleration benchmarks I tried:
- I removed the memory probes and only set memory logging to nvidia and tried with higher batch sizes but the benchmarking is still showing improved memory but the same train_runtime as running with the previous kernels. I had to run with the mistral model as the llama model with larger batch sizes was hitting OOM
   - GPU=1, batch_size=16, train_runtime=1803 (none)--> 1573 (foak-fast-kernels) -->1570 (foak-fast-kernels)-liger
   - GPU=2, batch_size=32, train_runtime=1792-->1563-->1560
   - GPU=4, batch_size=64, train_runtime=1863-->1632-->1629
   - GPU=4, batch_size=64, FT, train_runtime=80777-->78182-->72640
- Llama3-8b, 4 GPUs, max_seq_length=512, dataset_num_samples=10000
   - Lora, batch_size=32 (per_device=8) — 149 —> 131 --> 133
   - Lora, batch_size=64 (per_device=16) — 132 --> 115 --> 116
   - FT, batch_size=32 —  191 --> 172 --> 179
   - FT, batch_size=64 — 167 --> 149 --> 152

As you can see the train_runtime improves going from 0 kernels to fast kernels but from fast kernels to liger fused kernel showed minimal or no speedup as Fabian showed as well. 

As noted Fabian, I had to decrease the dataset size and max_seq_length in order to run full benchmarks which is what Aaron had done in the issue,

fabianlim:
@anhuong but isnt this a speedup for llama3? 

`GPU=4, batch_size=64, FT, train_runtime=80777-->78182-->72640`

We expect more speedup for larger vocabularly, in which case llama3 will see more benefit than mistral

anhuong:
Sorry that was memory metrics, wrote down the wrong info. I did see a small speedup in FT with batch_size=100 adding above

scripts/benchmarks/refs/a100_80gb_liger.csv
```diff
@@ -0,0 +1,121 @@
...
+,0.15,True,accelerated-peft-autogptq,2e-4,16,0.1,14776.0,9542673920.0,2261220352.0,TheBloke/Mistral-7B-v0.1-GPTQ,2,lora,2,16,q_proj k_proj v_proj o_proj,float16,0.9975294589996337,289.7083,1.381,0.345,2827.672
+,0.29,True,accelerated-peft-autogptq,2e-4,16,0.1,33739.0,26343190016.0,4337216000.0,TheBloke/Mistral-7B-v0.1-GPTQ,1,lora,8,16,q_proj k_proj v_proj o_proj,float16,0.97410005569458,955.3479,0.837,0.105,3429.955
+,0.29,True,accelerated-peft-autogptq,2e-4,16,0.1,22294.0,16188810752.0,2261416960.0,TheBloke/Mistral-7B-v0.1-GPTQ,2,lora,4,16,q_proj k_proj v_proj o_proj,float16,0.9795886325836182,509.3915,1.571,0.196,3216.387
+,0.15,True,accelerated-peft-autogptq-foak,2e-4,16,0.1,19715.0,13095119872.0,4336822784.0,TheBloke/Mistral-7B-v0.1-GPTQ,1,lora,4,16,q_proj k_proj v_proj o_proj,float16,1.0186691761016846,417.6284,0.958,0.239,3923.104
+,0.15,True,accelerated-peft-autogptq-foak,2e-4,16,0.1,14563.0,9326863872.0,2261220352.0,TheBloke/Mistral-7B-v0.1-GPTQ,2,lora,2,16,q_proj k_proj v_proj o_proj,float16,1.0090518665313721,223.4905,1.79,0.447,3665.48
```
fabianlim, line 91: I think we shouldnt create a different file for this, otherwise there will be repetition of these banches `accelerated-peft-autogptq-foak`
- I maybe refreshing the benches in #106, maybe we can these new benches on top

sample-configurations/foak-fast-kernels-sample-configuration.yaml
```diff
@@ -22,10 +22,13 @@ plugins:
       # - the FastQuantized version is all-or-nothing
 
       # fast loss triton kernels
-      fast_loss: True
+      fast_loss: False
```
fabianlim, line 25: the config must be properly installed in the benchmark utilty ,.we cannot do hardcoding like this otherwise all the other benches will be affected.

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/models/mixtral.py
```diff
@@ -93,6 +95,11 @@ def get_mp_rules(base_type):
                 "transformers.models.mixtral.modeling_mixtral",
             ),
         ),
+        ModelPatcherRule(
```
fabianlim, line 98: the better way is to do a swtich between the two rules `mixtral-cross-ent"` and `mixtral-fused-lce`, because they shouldnt be installed simultanosuly. If we know that we want fused, then only the fused rule should be installed.

plugins/fused-ops-and-kernels/configs/fast_kernels.yaml
```diff
@@ -22,4 +22,7 @@ training:
     fast_rms_layernorm: True
 
     # fast RoPE embedding triton kernels
-    fast_rope_embeddings: True
\ No newline at end of file
+    fast_rope_embeddings: True
+
+    # fused linear cross entropy loss
```
fabianlim, line 27: maybe but some more comments on how setting this will true will require fast_loss setting to be overrrden, or invalidated. or whatever we deciode

fabianlim:
@anhuong I did quite a few changes to this PR
- refactor: I moved the liger code to be under the `fused_ops` folder, since its more of that
- benches: I updated the `scenarios-liger.yaml` benches to better feel what it should be tested.
- configs: i changed `fast_loss` to now accept `True, False` or `fused_ce_liger`

### New Benches

image

image

anhuong:
Tested running tuning with fms_acceleration and fms_acceleration_foak from this branch and ran successfully.

Fine tuning on 1 epoch with config, note setting `"fast_kernels": ["fused_ce_liger", true, true]`:

<details>
<summary>JSON config:</summary>

```json
{
    "model_name_or_path": "/llama3/hf/8b_pre_trained/",
    "training_data_path": "/fmaas-integration-tests/tuning/input/twitter-complaints.json",
    "output_dir": "/fmaas-integration-tests/tuning/output/anhuong/fms_acceleration_benchmarking/llama3-8b_ft_twitter_from-fms-hf-tuning_liger",                                                                   
    "fast_kernels": ["fused_ce_liger", true, true],                                                       
    "save_strategy": "epoch",
    "num_train_epochs": 1.0,
    "per_device_train_batch_size": 4,
    "gradient_accumulation_steps": 1,
    "learning_rate": 1e-5,
    "response_template": "\n### Label:",
    "dataset_text_field": "output"
}
```

</details>

Logs showing llama-fused-lce being used
```sh
***** FMS AccelerationFramework *****
INFO:framework.py:***** FMS AccelerationFramework *****
Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.3.3.dev0.
INFO:framework.py:Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.3.3.dev0.
INFO:framework.py:***************** Module Forwards Patching *************
Rule: llama-fused-lce Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-fused-lce Module:                           Class: LlamaForCausalLM Num:  1
Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
INFO:framework.py:Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
```

Also tested running with triton cross entropy by setting `"fast_kernels": [true, true, true]` which ran as expected.

Running with the granite model I do see
```
***** FMS AccelerationFramework *****
INFO:framework.py:***** FMS AccelerationFramework *****
Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.3.3.dev0.
INFO:framework.py:Active Plugin: FastKernelsAccelerationPlugin. Python package: fms_acceleration_foak. Version: 0.3.3.dev0.
INFO:framework.py:***************** Module Forwards Patching *************
Rule: granite-fused-lce Module:                           Class: GraniteForCausalLM Num:  1
INFO:framework.py:Rule: granite-fused-lce Module:                           Class: GraniteForCausalLM Num:  1
Rule: granite-rms     Module: input_layernorm           Class: GraniteRMSNorm  Num: 40
INFO:framework.py:Rule: granite-rms     Module: input_layernorm           Class: GraniteRMSNorm  Num: 40
Rule: granite-rms     Module: model                     Class: GraniteRMSNorm  Num:  1
INFO:framework.py:Rule: granite-rms     Module: model                     Class: GraniteRMSNorm  Num:  1
Rule: granite-rms     Module: post_attention_layernorm  Class: GraniteRMSNorm  Num: 40
INFO:framework.py:Rule: granite-rms     Module: post_attention_layernorm  Class: GraniteRMSNorm  Num: 40
Rule: granite-rope    Module:                           Class: GraniteForCausalLM Num:  1
INFO:framework.py:Rule: granite-rope    Module:                           Class: GraniteForCausalLM Num:  1
```
@fabianlim is there anything else I should check to see if it runs with the granite model?

fabianlim:
@anhuong can we just do a quick check with granite?. Also can we update how to run these [new liger benches ](https://github.com/foundation-model-stack/fms-acceleration/pull/93#issuecomment-2481736149) in the [README.md](https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/fused-ops-and-kernels/README.md)

anhuong:
> Can i confirm that when you run with "fast_kernels": [true, true, true] the fused_ce does NOT get activated?

That is correct, I'll update the above to include the log outputted

fabianlim (approved):
LGTM

Issue #112, opened 2024-11-20
Enable GPTQModel to handle GraniteMoeParallelExperts


fabianlim:
Granite MoE uses a [3D tensor](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granitemoe/modeling_granitemoe.py#L276) to hold the expert weights, so GPTQModel does not work out of the box.

There are two options
1. module swap `GraniteMoeParallelExperts` to hold a ModuleList of Linears, then AutoGPTQ will be able to detect them and replace them with `QuantLinears`
2. write a custom [gptq](https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py#L52) module that handles the `GraniteMoeParallelExperts case 

Either of the two approaches will solve both quant + inference paths. Option 1 should be easier than Option 2, but in some sense Option 2 should be more proper.

When doing option 2 we should be reusing code from the original [gptq](https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py#L52). 
- also it should be written generally, to not just only handle this particular `GraniteMoeParallelExperts` instance, but all cases with 3D tensors

Abhishek-TAMU:
[Test script](https://ibm.box.com/s/qooh8x7emb7yowgt8yf3502bctyu5pbp) used to test the draft PR to quantize `Granite3.0-3b-MoE` model (`/ibm-dmf-lakehouse/models/base_training/shared/granite-3.0-3b-a800m-base/r240924a`).

As per Praveen, this work could be further continued later in Q2 of 2025 when new Granite MoE 120b model will come out.

PR #122, opened 2025-01-28
[WIP] Enable GPTQModel to handle GraniteMoeParallelExperts

Abhishek-TAMU:
ISSUE: https://github.com/foundation-model-stack/fms-acceleration/issues/112

Changes made to enable `GraniteMoeParallelExperts` layer in GPTQModel :

1- `find_layers` function in `plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/utils/model.py`
2- `class GPTQ` : `plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py`
- `init` and `add_batch` function 
- TODO: `fasterquant` function


plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/base.py
```diff
@@ -558,7 +558,7 @@ def save_quantized(
         self.quantize_config.meta_set_versionable(
             key=META_FIELD_QUANTIZER,
             value=META_QUANTIZER_GPTQMODEL,
-            version=__version__,
+            version="1.0.0",
```
fabianlim, line 561: why does this need to be changed?

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
```diff
@@ -25,46 +26,113 @@ class GPTQ:
...
-            inp = inp.flatten(1)
-        self.H *= self.nsamples / (self.nsamples + tmp)
-        self.nsamples += tmp
-        # inp = inp.float()
-        inp = math.sqrt(2 / self.nsamples) * inp.float()
-        # self.H += 2 / self.nsamples * inp.matmul(inp.t())
-        self.H += inp.matmul(inp.t())
+        # Update entire H_list and nsamples_list
+        if not self.is_moe:
+            # print("INSIDE ADD_BATCH FOR 2D")
```
fabianlim, line 78: this looks abit overcomplicated. I think lets not have an `else` clause and assume that is for "inside an MoE". Im worried that this logic is rather fragile and could break.

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
```diff
@@ -74,137 +142,162 @@ def fasterquant(
...
-        return scale, zero, g_idx, duration, avg_loss
+        # For MoE model
+            # Loop over each expert param and quantize it separately 
+            t_start = time.time()
+            scale_list = []
+            zero_list = []
+            gidx_list = []
+            loss_list = []
+
+            for i in range(self.num_experts):
```
fabianlim, line 287: I feel there is quite a lot of complication but in actuality the MoE design is quite simple. Assuming you dont quantize the router, then the only thing that needs to be quantized are the experts
- so right now the experts are put into `ParameterList` and so all this custom logic to loop over stuff needs to be written, which is not ideal
- however, maybe you can consider a different design. For example, if you create a `ModuleList` instead and put each expert as a 2D Linear module inside that list, then I think perhaps alot of this custom logic does not need to be written

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
Abhishek-TAMU: > I think lets not have an else clause and assume that is for "inside an MoE"

I didn't understand this. The current logic of `add_batch` for 2D deals with `self.H` but for 3D we have `self.H_list`, hence I assume when registering forward hook, all the H's in `self.H_list` would be updated. Though not sure on if H's are supposed to be updated equally for all experts, which I have done here.

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
fabianlim: I just checked and this is what `Mixtral` is doing

```
MixtralDecoderLayer(
  (self_attn): MixtralSdpaAttention(
    (q_proj): Linear(in_features=4096, out_features=4096, bias=False)
    (k_proj): Linear(in_features=4096, out_features=1024, bias=False)
    (v_proj): Linear(in_features=4096, out_features=1024, bias=False)
    (o_proj): Linear(in_features=4096, out_features=4096, bias=False)
    (rotary_emb): MixtralRotaryEmbedding()
  )
  (block_sparse_moe): MixtralSparseMoeBlock(
    (gate): Linear(in_features=4096, out_features=8, bias=False)
    (experts): ModuleList(
      (0-7): 8 x MixtralBlockSparseTop2MLP(
        (w1): Linear(in_features=4096, out_features=14336, bias=False)
        (w2): Linear(in_features=14336, out_features=4096, bias=False)
        (w3): Linear(in_features=4096, out_features=14336, bias=False)
        (act_fn): SiLU()
      )
    )
  )
  (input_layernorm): MixtralRMSNorm((4096,), eps=1e-05)
  (post_attention_layernorm): MixtralRMSNorm((4096,), eps=1e-05)
)

```

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
```diff
@@ -0,0 +1,30 @@
...
+###############################################################################
+# Local
+from .base import BaseGPTQModel
+
+
+class GraniteMoeGPTQ(BaseGPTQModel):
+    base_modules = ["model.embed_tokens", "model.norm"]
+
+    layers_node = "model.layers"
+    layer_type = "GraniteMoeDecoderLayer"
```
fabianlim, line 24: I suggest you add some simple key to inform the format of `input_linear` and `output_linear`, that these are 3D tensors.

Also in the granitemoe case, another compilation is that `input_linear` fuses `w1` and `w3`. it might be ok for a first cut just to leave them as fused.

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
```diff
@@ -0,0 +1,30 @@
...
+    layer_modules = [
+        ["self_attn.k_proj", "self_attn.v_proj", "self_attn.q_proj"],
+        ["self_attn.o_proj"],
+        ["block_sparse_moe.input_linear", "block_sparse_moe.output_linear"],
```
fabianlim, line 28: reference `MixtralGPTQ`, you will see that they split up `w1`+ `w3` and `w2`, which means we should split `block_sparse_moe.input_linear` and `"block_sparse_moe.output_linear"`

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/utils/model.py
```diff
@@ -101,11 +102,19 @@ def nested_move_to(v, device):
 
 
 def find_layers(module, layers=None, name=""):
+    # print("1- INSIDE find_layers module", module)
     if not layers:
-        layers = [transformers.pytorch_utils.Conv1D, nn.Conv2d, nn.Linear]
+        # Can add MOE.GraniteMoeRMSNorm here if want to include Linear Norm layer ["input_layernorm", "post_attention_layernorm"]
+        # MOE.GraniteMoeParallelExperts is torch.nn.Module for layer ["block_sparse_moe.input_linear", "block_sparse_moe.output_linear"]
+        layers = [transformers.pytorch_utils.Conv1D, nn.Conv2d, nn.Linear, MOE.GraniteMoeParallelExperts]
```
fabianlim, line 109: if you can transform `module` into `ModuleList` , you dont need to modify this function

fabianlim: that means `input_linear` -> with a `ModuleList`, and then we emulate the behaviors

means that `input_linear(input)` gives a 3D tensor, then even when its `ModuleList` we try to emuilate it and make it also give a 3D tensor

so maybe a direct replacement of `input_linear` is not possible, but replace it with a class inheritied from `ModuleList`

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
```diff
@@ -25,46 +26,113 @@ class GPTQ:
     def __init__(self, layer):
```
fabianlim, line 26: if layer has been modified such that all `3D` tensors are replaced with module lists, then no change in this function is needed

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
```diff
@@ -74,137 +142,162 @@ def fasterquant(
...
-
-                err1 = (w - q) / d
-                W1[:, i:] -= err1.unsqueeze(1).matmul(Hinv1[i, i:].unsqueeze(0))
-                Err1[:, i] = err1
-
-            Q[:, i1:i2] = Q1
-            Losses[:, i1:i2] = Losses1 / 2
-
-            W[:, i2:] -= Err1.matmul(Hinv[i1:i2, i2:])
+        if not self.is_moe:
```
fabianlim, line 145: same.. if module list replacement is done, then no logic change is needed

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
fabianlim: so basically the simple key needs to know what do look for to convert it to 3D tensor

```python
class GraniteMoeGPTQ(BaseGPTQModel):
    
    convert3dToModuleList = ["block_sparse_moe.input_linear", "block_sparse_moe.output_linear"]

    layer_modules = [

        [
             "block_sparse_moe.input_linear.0.weight",
              "block_sparse_moe.input_linear.1.weight",
              ...
        ], [
             "block_sparse_moe.output_linear.0.weight",
              "block_sparse_moe.output_linear.1.weight",
              ...
        ]
    ]
```

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/quantization/gptq.py
fabianlim: yes lets not introduce this logic, because we know this function already works for `mixtral`, which is an MoE. So the design of this function does not really need introduction of `H_list` semantics, it is a product of our earlier decision, to use `ParameterList`, which in hindsight may not be such a good idea now given that so many extensive logic changes are required in these internals

Abhishek-TAMU:
@fabianlim Based on your comments I made a incremental change of using ModuleList with 2D `nn.Linear` layers to store 3D weight of expert layer. Just wanted to ask is this the direction we want to go forward with ?

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/utils/model.py
```diff
@@ -100,14 +120,40 @@ def nested_move_to(v, device):
...
+        module = ThreeDTensorModuleList([
+            nn.Linear(input_size, output_size, bias=False) for _ in range(num_experts)
+        ])
+
+    return module
+
+
 def find_layers(module, layers=None, name=""):
+    # print("1- INSIDE find_layers module", module)
+    module = check3DTensor(module, name)
```
fabianlim, line 138: this logic can be done outside of this function. 
1. load model
2. replace all the 3d layers
3. then call `GPTQModel`

fabianlim, line 138: what is missing here is the logic for having the model recipe needs to pass in the `convert3dToModuleList` How do you plan to handle this?

One alternative is to handle the module swap completely outside of this function, then you dont need this logic

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/utils/model.py
```diff
@@ -100,14 +120,40 @@ def nested_move_to(v, device):
         return v
 
 
+def check3DTensor(module, name, convert3dToModuleList=["block_sparse_moe.input_linear", "block_sparse_moe.output_linear"]):
+    if convert3dToModuleList and name in convert3dToModuleList:
+        # print("INSIDE check3DTensor module, name, convert3dToModuleList", module, name, convert3dToModuleList)
+        num_experts = module.num_experts
```
fabianlim, line 126: same comment as above, its written very specifically for an MoE, and also it assumes attributes like `num_experts` which may not exist generally.

If its too difficult to write a robust `check` function, then i suggest somehow to write a hook mechanism to take in `convert3dToModuleList` from `GraniteMoeGPTQ` and just do all the module swaps outside first in a prescibed manner

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###############################################################################
+# Local
+from .base import BaseGPTQModel
+
+
+class GraniteMoeGPTQ(BaseGPTQModel):
+    base_modules = ["model.embed_tokens", "model.norm"]
+    convert3dToModuleList = ["block_sparse_moe.input_linear", "block_sparse_moe.output_linear"]
```
fabianlim, line 22: `convert3dToModuleList` has to be part of the `BaseGPTQModel` design, so it needs to exist in the parent class

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+###############################################################################
+# Third Party
+import torch
+
+# Local
+from .base import BaseGPTQModel
+
+
+def new_forward(self, inputs, expert_size):
```
Abhishek-TAMU, line 23: I was testing by defining this `new_forward` method here. But as you said if we are expecting it from users then does user pass this function ? Is it in GPTQModel.pretrained() function [here](https://github.com/Abhishek-TAMU/fms-acceleration/blob/f98e85dadfc4c6d6f61c4f18aedc9ea1c165ac5d/plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/auto.py#L70) and then from there passing it forward to `BaseGPTQModel.init()` ?

Abhishek-TAMU (commented):
@fabianlim I was testing with your code and it is working now (quantizing `block_sparse_moe` layer). I have a question though on `new_forward` method I used for testing. Asked question below.

Abhishek-TAMU:
After `block_sparse_layer` along with other layers are successfully quantized, there is an error in packing the model part:

image

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/granitemoe.py
fabianlim: No. the user is the one who writes the new model, in this case, `GraniteMoeGPTQ`, 
- so while user is writing `GraniteMoeGPTQ`, user will then configure all the forwards to be overwritten in the `update_forwards` class member
So how you have done it now is jsut fine

Issue #98, opened 2024-10-29
FOAK Cross Entropy Loss Will Not Work with New Loss Functions After Transformers 4.46


fabianlim:
There has been significant refactoring of the loss functions for `transformers 4.46`, that will render the cross entropy patching ineffective. Need to have a different `ModelPatcherRule` for the new transformers version. CC: @anhuong 

https://github.com/huggingface/transformers/pull/34191

So now there are 3 possiblities
1. `custom_loss_function` is passed into `Trainer`
2. model has migrated to the `custom_loss_function` API
3. model has not migrated (like Granite now)

For 3. This is the easy one, because it means no code changes 

For 1. Im thinking we do not patch anything, because if a user wants to do this, we cant control what loss function they use

For 2. In this case we want to patch[ `fixed_cross_entropy` ](https://github.com/huggingface/transformers/blob/dbbc3ce85ff2d3432467cbc5dc39165ec8275363/src/transformers/loss/loss_utils.py#L24), but this should be done on a per-model basis. So we need to somehow have the model instantiate the loss function, e.g., `ForCausalLMLoss`, and only patch `fixed_cross_entropy` during this instantiation process, and put it back to original after it is done

PR #123, opened 2025-02-05
fix: cross entropy for transformers>4.45

anhuong:
- Checks for transformers version and creates new custom loss function for llama, granite, mistral, and mixtral models. 
- Adds shard_checkpoint function from transformers as it is missing in later versions

Tested and saw cross-entropy switching based on transformers version correctly. Benchmark singular images posted below here is a single image comparing the previous benchmark to the new one. Overall the train_loss is on par, memory use is on par/a little higher, and train_tokens_per_second is larger/slower.

Screenshot 2025-02-07 at 12 04 35 PM

closes: #98


plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
```diff
@@ -290,3 +291,55 @@ def forward(self, input, target):
...
+    shift_labels = labels[..., 1:].contiguous()
+
+    # Flatten the tokens
+    shift_logits = shift_logits.view(-1, vocab_size)
+    shift_labels = shift_labels.view(-1)
+    # Enable model parallelism
+    shift_labels = shift_labels.to(shift_logits.device)
+
+    reduction = "sum" if num_items_in_batch is not None else "mean"
+    assert ignore_index == -100, "FastForCausalLMLoss currently supports only hardcoded ignore index -100."
```
anhuong, line 316: What is -100 ignore_index, I see that ignore_index is the target value that is ignored and does not contribute to the input gradient, but for CausalLMLoss what is at index -100?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
```diff
@@ -290,3 +291,55 @@ def forward(self, input, target):
...
+    loss = Fast_CrossEntropyLoss.apply(
```
anhuong, line 317: Can you describe the difference between `Fast_CrossEntropyLoss` and `FastCrossEntropyLoss`?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
```diff
@@ -290,3 +291,55 @@ def forward(self, input, target):
         )
         n_items = torch.count_nonzero(target != -100)
         return loss.sum() / n_items
+
+
+# added by flim@sg.ibm.com
+
+# adapted from transformers.loss.loss_utils.ForCausalLMLoss
+def FastForCausalLMLoss(
```
anhuong, line 299: Would we need to create a similar FastForCausalLMLoss for liger kernel cross entropy?

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/models/granite.py
```diff
@@ -122,16 +123,25 @@ def get_mp_rules(base_type: str, config: PretrainedConfig = None):
...
-                FastCrossEntropyLoss,
-                "transformers.models.granite.modeling_granite",
-            ),
-        ),
+        *[
+            ModelPatcherRule(
+                rule_id="granite-custom-loss",
+                trigger=ModelPatcherTrigger(
+                    check=replace_custom_loss_when_triggered(
+                        GraniteForCausalLM, custom_loss_type="granite-custom-loss"
```
anhuong, line 131: Thoughts on calling this `granite-custom-crossent-loss` instead to be specific that the custom loss is for cross entropy

anhuong (commented):
Had some questions when going through the code

anhuong:
I also noticed in the new transformers version there is a lot of slowness after loading on the checkpoint on log line:
```
The new embeddings will be initialized from a multivariate normal distribution that has old embeddings' mean and covariance. As described in this article: https://nlp.stanford.edu/~johnhew/vocab-expansion.html. To disable this, use `mean_resizing=False`
```

Is this something we want to do? Is it expected that loading the embeddings will take a long time to run?

fabianlim:
@anhuong the slownesss i feel its due to the recent changes in `fms-hf-tuning`, and is there because now we will resize the embedding layer if there is special tokens. Previously we didnt use to do that.

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/models/granite.py
fabianlim: I feel `custom-loss` is ok, because it mostly refers to that we are using the new custom loss feature.

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
fabianlim: `-100` is used extensively throughout HF, while they provide some means for user to change it, almostly nobody will bother to change it

It is the `label` that is at -100. For a label with that value, we will ignore that token's contribution to the loss

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
fabianlim: - `Fast_CrossEntropyLoss` is the autograd function. we inherit this from unsloth
- `FastCrossEntropyLoss` is a specialization of `torch.nn.CrossEntropyLoss` that serves as a convinienced, implemted using `Fast_CrossEntropyLoss`

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
fabianlim: yes I think we will have a new function for liger cross entropy with the same API. then its a plug and play. But it should be used only if the transformer versioin is advanced

anhuong:
Able to verify enabling the correct cross-entropy is triggeed based on transformers version for granite, llama, mistral, and mixtral models.

With transformers==4.48
```sh
Rule: llama-custom-loss Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-custom-loss Module:                           Class: LlamaForCausalLM Num:  1
Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
INFO:framework.py:Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
```

With transformers=4.45
```sh
Rule: llama-cross-ent Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-cross-ent Module:                           Class: LlamaForCausalLM Num:  1
Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: input_layernorm           Class: LlamaRMSNorm    Num: 32
Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
INFO:framework.py:Rule: llama-rms       Module: model                     Class: LlamaRMSNorm    Num:  1
Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
INFO:framework.py:Rule: llama-rms       Module: post_attention_layernorm  Class: LlamaRMSNorm    Num: 32
Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
INFO:framework.py:Rule: llama-rope      Module:                           Class: LlamaForCausalLM Num:  1
```

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/base.py
```diff
@@ -766,6 +769,106 @@ def save_quantized(
         quantize_config.model_file_base_name = model_base_name
         quantize_config.save_pretrained(save_dir)
 
+    # added by anh.uong@ibm.com
+    # adapted from transformers.modeling_utils.shard_checkpoint
+    # from transformers v4.46, removed in later versions
+    # TODO: split_torch_state_dict_into_shards from huggingface_hub library
+    def shard_checkpoint(
```
anhuong, line 776: After transformers v4.46, this method no longer exists in in transformers so I copied it in here to start. The new method to migrate to as per the [warning message in the original function ](https://github.com/huggingface/transformers/blob/v4.46.0/src/transformers/modeling_utils.py#L400-L402) says to migrate to `split_torch_state_dict_into_shards` as noted in the TODO item here. This method was similar but requires more investigation on the difference - https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/serialization/_torch.py#L302

scripts/benchmarks/refs/a100_80gb_update_cross_entropy_and_deps.csv
```diff
@@ -0,0 +1,89 @@
+bf16,epoch,fp16,framework_config,learning_rate,lora_alpha,lora_dropout,mem_nvidia_mem_reserved,mem_peak_torch_mem_alloc_in_bytes,mem_torch_mem_alloc_in_bytes,model_name_or_path,num_gpus,peft_method,per_device_train_batch_size,r,target_modules,torch_dtype,train_loss,train_runtime,train_samples_per_second,train_steps_per_second,train_tokens_per_second
+,0.07,,none,2.00E-05,,,15116,11267745280,6770300416,bigcode/gpt_bigcode-santacoder,1,,4,,,bfloat16,2.33703125,47.6604,8.393,2.098,17188.262
```
anhuong, line 2: I ran the benchmarks and found the failure in auto-gptq that I commented on the fix above. Rerunning benchmarks for auto-gptq. I also ran benchmarks for granite3.1 model but for comparison I had to run against the granite-gptcode model. Do we want to update to update to the granite3.1 model? It did run successfully with it as well.

Here are the charts showing the comparison against a100_80gb.csv without including the auto-gptq failed runs shows the train_loss is on par, memory use is a little improved, and train_tokens_per_second is larger/slower
compare-mem_peak_torch_mem_alloc_in_bytes-crossent-full-match-more
compare-mem_torch_mem_alloc_in_bytes-crossent-full-match-more
compare-train_loss-crossent-full-match-more
compare-train_tokens_per_second-crossent-full-match0more

plugins/fused-ops-and-kernels/src/fms_acceleration_foak/kernels/unsloth/cross_entropy_loss.py
anhuong: I was thinking of adding this change in a separate PR

scripts/benchmarks/refs/a100_80gb_update_cross_entropy_and_deps.csv
fabianlim: these look quite decent

plugins/accelerated-peft/src/fms_acceleration_peft/gptqmodel/models/base.py
fabianlim: ok this is fine for now

fabianlim (approved):
LGTM the benches look good.

One question is that the new benches you ran. rather than have it in a seperate file, it is better to replace the previous `a100_80gb` and update the requirements, but did you run all the cases or only a subset?

anhuong:
I will rename the benchmarks and requirements, I had run all of the benchmarks except for auto-gptq due to the error that came up so I ran those separately after the fix. I also did not run the baseline-bnb but running now separately, what is the purpose of this benchmark? updated the image in the description and will include individual images here. I will add all of them to the benchmark. I updated the above description with the summary results that include auto-gptq, Here are the individual images. Overall they continue to look good, the only outlier identified was


framework_config | peft_method | model_name_or_path | num_gpus | per_device_train_batch_size | reference | metric | new
-- | -- | -- | -- | -- | -- | -- | --
accelerated-peft-bnb | lora | bigcode/gpt_bigcode-santacoder | 2 | 2 | 6840.355 | train_tokens_per_second | 8548.922
accelerated-peft-bnb-foak | lora | bigcode/gpt_bigcode-santacoder | 2 | 2 | 10345.932 | train_tokens_per_second | 11994.044

anhuong:
I have replaced the benchmark and requirements with my full runs

fabianlim:
> I also did not run the baseline-bnb but running now separately, what is the purpose of this benchmark? updated the image in the description and will include individual images here.
the purpose of this is to have a baseline so we can compare the accelerations. The baseline could change also due to different transformer versions

anhuong:
Makes sense, the benchmark I added has the complete benchmark including the baseline. It matches the original. With this, I will merge in this change

fabianlim:
sounds good!

PR #127, opened 2025-02-21
fix(deps): Copy KHD imports into `scattermoe_utils`

willmj:
Scatter MoE has been using [kernel hyperdrive](https://github.com/fabianlim/kernel-hyperdrive) as an external dependency since it's not a pypi library, which makes installation more difficult. Therefore we copy the functions we depend on into `scattermoe_utils.khd` to remove this dependency.


plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_utils/khd/kernels/ops/compileable_ops.py
```diff
@@ -0,0 +1,349 @@
+# Third Party
```
fabianlim, line 1: haev a line here that references cute kernels would be good.

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_utils/khd/kernels/ops/__init__.py
```diff
@@ -0,0 +1,310 @@
+# Copyright The FMS HF Tuning Authors
+# Copyright 2024 Databricks
```
fabianlim, line 2: this file is not from databricks, it is from cute kernels

fabianlim (approved):
LGTM just some minor changes

PR #129, opened 2025-02-23
fix: Allow for multi node training for accelerated moe

kmehant:
Current implementation uses global rank of the process to prepare the device index which would not work in a multi node setting. Therefore, we would need to use local rank since devices are not continuously indexed across the nodes.

### DIFF
--- a/plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
+++ b/plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
@@ -69,7 +69,8 @@ class ScatterMoEAccelerationPlugin(AccelerationPlugin):
         rank, world_size = 0, 1
         if torch.distributed.is_initialized():
             world_size = torch.distributed.get_world_size()
-            rank = torch.distributed.get_rank()
+            # we do not need to use the fallback as this is wrapped in an `is_initialized` block
+            rank = torch.distributed.get_node_local_rank()
 
         if not hasattr(model.config, "name_or_path") or not model.config.name_or_path:
             raise ValueError(


fabianlim:
@kmehant i understand the fix, but can you update the description for record keeping purposes

kmehant:
https://github.com/foundation-model-stack/fms-acceleration/pull/129#issuecomment-2677333723

Apologies for missing that, have added it.

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
```diff
@@ -65,7 +66,7 @@ def augmentation(
         rank, world_size = 0, 1
         if torch.distributed.is_initialized():
             world_size = torch.distributed.get_world_size()
-            rank = torch.distributed.get_rank()
+            rank = int(os.environ["LOCAL_RANK"])
```
fabianlim, line 69: can we make it consistent and follow the new [style](https://github.com/pytorch/pytorch/pull/126737). 
```suggestion
            # we do not need to use the fallback as this is wrapped in an `is_initialized` block   
            rank = torch.distributed.get_node_local_rank()
```

fabianlim (approved):
LGTM but one suggestion

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
kmehant: Have included this suggestion thanks.

kmehant:
@fabianlim requesting your merge.

fabianlim:
@kmehant lets have @willmj look at it first

PR #133, opened 2025-03-06
chore: Move dcp to safetensors to a wrapper function for downstream usage

kmehant:
Add a wrapper function to wrap complete dcp to safetensor logic in order to enable easy downstream usage through this function without needing to redundantly repeat the same set of steps.


fabianlim:
@kmehant actually this design goes agains the ADR. we purposedly did not want fms-acceleration to modify the trainer. I feel the callback should be better installed in fms-hf-tuning

kmehant:
@fabianlim or @willmj Shall we merge this?

PR #136, opened 2025-03-12
feat: Allow for MoE kernels (Scatter MoE) irrespective of use of EP

kmehant:
Today, we would need to use ep_degree > 0 to trigger use of MoE kernels + EP. This PR adds a way to trigger use of MoE Kernels without EP by setting ep_degree = 0


plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -118,6 +118,11 @@ def prepare_scattermoe(
     # pylint: disable=import-outside-toplevel
     from .scattermoe import ScatterMoE
 
+    no_ep_no_replication = False
```
fabianlim, line 121: `no_ep_no_replication` is really quite a bad name for a variable.

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -131,6 +136,15 @@ def prepare_scattermoe(
     # current rank of the device
     device = torch.device(f"{device_type}:{rank}")
 
+    # NOTE: fsdp_cpu_ram_efficient_loading is not supported for EP activated cases
+    fsdp_cpu_ram_efficient_loading = False
```
fabianlim, line 140: its better to use this [function](https://github.com/huggingface/transformers/blob/7426d02ea8396aef6206fdbcdd7ae92f96a468b4/src/transformers/modeling_utils.py#L159-L165)

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -144,6 +158,8 @@ def prepare_scattermoe(
     expert_name = expert_name.split("|")
 
     rep_size = world_size // ep_degree
+    if no_ep_no_replication:
```
fabianlim, line 161: for better readiblity, should have this within the conditions starting from 163 below

kmehant:
@fabianlim The PR is ready, let me know if I have missed something. I have update the PR description with summary of changes. Thanks

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
```diff
@@ -91,23 +91,25 @@ def get_callbacks_and_ready_for_train(
...
-
-            # call this to patch the HF save and load functions to be able
-            # to save DTensors propery
-            patch_huggingface_save_and_load_for_dtensors()
-
-            # call this to patch torch optim to not use
-            # foreach for dtensors
-            patch_torch_optim_foreach_to_not_apply_to_dtensors()
+            # When EP is not enabled we want to shard the experts using FSDP
+            if self._ep_degree != 0:
```
fabianlim, line 95: @kmehant im not a fan of doing `ep_degree == 0`. I feel if you want to merge this it should be addressed. 

best is to have an external control, the below is a bad example. so can you think of a better one and propose. 

```
training.moe.scattermoe.disable_ep # default is False
```

This is bad because when people set `ep_degree=1`, they might think that ep is already disabled. So we need to think of something else more intuitive.

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -131,6 +139,16 @@ def prepare_scattermoe(
     # current rank of the device
     device = torch.device(f"{device_type}:{rank}")
 
+    # NOTE: fsdp_cpu_ram_efficient_loading is not supported for EP activated cases
+    fsdp_cpu_ram_efficient_loading = is_fsdp_enabled()
```
fabianlim, line 143: i feel there is no need to define this variable.. just call `is_fsdp_enabled` directly

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -118,6 +118,14 @@ def prepare_scattermoe(
     # pylint: disable=import-outside-toplevel
     from .scattermoe import ScatterMoE
 
+    ep_disabled = False
```
fabianlim, line 121: I think you can make this as an input to `prepare_scattermoe`, it can be called the same as whatevern `training.moe.scattermoe.disable_ep` is going to be called.

Then we do not need this weird if `ep_degree == 0` logic, then set `ep_degree = 1` logic

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
```diff
@@ -131,6 +139,16 @@ def prepare_scattermoe(
     # current rank of the device
     device = torch.device(f"{device_type}:{rank}")
 
+    # NOTE: fsdp_cpu_ram_efficient_loading is not supported for EP activated cases
+    fsdp_cpu_ram_efficient_loading = is_fsdp_enabled()
+
+    if ep_disabled:
```
fabianlim, line 145: we should set the device of rank 0 only if `is_fsdp_enabled() == True`

fabianlim (changes requested):

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
kmehant: We can add a flag calling `enable_fsdp_shard_experts` or `enable_fsdp_shard_moe` which is only used when `ep_degree==1`. WDYT?

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
kmehant: Done

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
fabianlim: I dont understand what you mean by flag, do you mean?

```
training.moe.scattermoe. enable_fsdp_shard_experts
```

If that is the case, then no, 
1. there is no FSDP in side `prepare_scattermoe`, so we shouldnt have the word "FSDP" inside there
2. we need something that discribes disabling behavior, because when it is set, then we basically do nothing. We should have have any semantic to suggest that we are doing something as `enable_fsdp_shard_experts` would be implying

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
fabianlim: you only set device == cpu only in the case when 
1. `is_fsdp_enabled()` == True, which will be the case when `low_cpu_mem_usage` is enabled
2. only in the case when rank == 0

Correct? If `is_fsdp_enabled()` == False, you shouldnt be setting the device to cpu

fabianlim: `is_fsdp_enabled() == True` implies low_cpu_mem_mode is [turned on](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L159-L164).

fabianlim (changes requested):

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
kmehant: @fabianlim Should we just enable FSDP sharding when `ep_degree=1`? Since users think EP is disabled while using FSDP would also mean experts would be now FSDP sharded?

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe_prepare.py
kmehant: > Correct? If is_fsdp_enabled() == False, you shouldnt be setting the device to cpu

I was thinking we load the MoE module on to CPU for no EP case even when low_cpu_mem_usage is off. This aligns with we loading the model to cpu during `from_pretrained()` before passing it to the trainer. However, I am okay making the change you suggested

plugins/accelerated-moe/src/fms_acceleration_moe/framework_plugin_scattermoe.py
```diff
@@ -42,12 +42,31 @@ def __init__(self, configurations: Dict[str, Dict]):
...
+        # Scatter MoE kernels.
+        #
+        # ep_degree==1 and disable_distributed is True mean use of Scatter MoE
+        # kernels + distribution deferred to top level distribution protocol (e.g. FSDP).
+        #
+        # ep_degree>1 and disabled_distributed is False mean enabling expert parallel
+        # and Scatter MoE Kernels.
+        #
+        # ep_degree>1 and disable_distributed is True errors out.
+
```
fabianlim, line 59: @willmj are you ok with the flag name `disable_distributed`?

```suggestion
        # If disable_distributed==False, the moe plugin handles sharding / replication,
        # otherwise user will need handle this manually (e.g., using FSDP)
        #
        # ep_degree=1 (default):
        # - disable_distributed=False (default) means
        # experts are replicated while using ScatterMoE kernels.
        # - disable_distributed=True means no replication (please use
        #    own training framework) 
        #
        # ep_degree > 1:
        # - disabled_distributed=False (default) means expert sharding with 
        # Scatter MoE Kernels.
        # disable_distributed=True cannot be set in this case; errors out.

```

fabianlim (approved):
i approve. @willmj can you just look through it.. and if ok you can merge this

willmj (approved):
Looks good to me. Just to confirm, QuantLinear works on 0.14.0 but doesn't work on 0.15.0? fms-hf-tuning is currently restricted to peft<0.14.0

kmehant:
https://github.com/foundation-model-stack/fms-acceleration/pull/136#pullrequestreview-2732981175

@willmj 
I had to arrive at this change, since the lint was failing (which is independent of fms-hf-tuning) since the `framework` module was installing latest possible peft package.  `QuantLinear` seem to be refactored from 0.15.0 which I think should be addressed separately. Thanks

willmj:
Thanks Mehant!

PR #140, opened 2025-04-14
fix: use element_size

kmehant:
`dtype_byte_size` is no more supported in transformers due to its inaccuracies - https://github.com/huggingface/transformers/pull/37144 and related issues.

We should move to torch native solution `element_size()`.


willmj (approved):
Looks good to me

fabianlim (approved):
yes looks good.

PR #139, opened 2025-04-09
feat: lora for accelerated MoE v1 - router only

willmj:
Building on #138 to add peft configs to fast moe augmentation so lora config is passed into `prepare_scattermoe`. Updating checkpoint utility functions to handle lora state dict.
Restrictions:

- lora_config.r must be >= 16
- Must be using FSDP, since the scatteredExperts weights are not supported by peft's LoRA tuning, the overwritten FSDP save and load functions must be utilized here.
- Loading from a default lora adapter config may not work here, intended use-case is to run tuning from base model.
- vLLM inference cannot load custom ScatteredExperts, so lora tuning only tunes the `router.layer`, not `w1`, `w2`, `w3`

Here are logs of the transformation in `checkpoint_utils` before and after the `recover_original_state_dict_from_checkpoint` [function](https://github.com/foundation-model-stack/fms-acceleration/blob/a5bdea2848e269f80e24e200f96e86a1e32caf42/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py#L323):
[scattermoe-router-lora.log](https://github.com/user-attachments/files/19673060/scattermoe-router-lora.log)

PR #141, opened 2025-04-17
feat: lora for accelerated MoE - limited

willmj:
Limited version of #139 without support for router or expert layers

### DIFF
--- a/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
+++ b/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
@@ -592,7 +592,9 @@ def recover_safetensors_from_dcp(
         new_state_dict[name] = param
 
     # recover the original state dict
-    state_dict = recover_original_state_dict_from_checkpoint(new_state_dict, _name_or_path)
+    state_dict = recover_original_state_dict_from_checkpoint(
+        new_state_dict, _name_or_path
+    )
 
     # save it as a safetensors file
     save_sharded_safetensors(
--- a/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
+++ b/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
@@ -540,6 +540,19 @@ def recover_safetensors_from_dcp(
     # get the state_dict
     state_dict = loader(checkpoint_dir)
 
+    new_state_dict = {}
+    for name, param in state_dict.items():
+        if "base_model.model." in name:
+            name = name.replace("base_model.model.", "", 1)
+        if "default." in name:
+            name = name.replace("default.", "", 1)
+        new_state_dict[name] = param
+
+    # recover the original state dict
+    state_dict = recover_original_state_dict_from_checkpoint(
+        new_state_dict, _name_or_path
+    )
+
     # recover the original state dict
     state_dict = recover_original_state_dict_from_checkpoint(state_dict, _name_or_path)
 
--- a/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
+++ b/plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
@@ -580,11 +580,15 @@ def recover_safetensors_from_dcp(
     # get the state_dict
     state_dict = loader(checkpoint_dir)
 
+    # filter out additional names created by lora tuning
+    # create switch based on state dict for future use
     new_state_dict = {}
     lora = False
     for name, param in state_dict.items():
+        # if lora weight, set lora switch to true
         if "lora_A" in name or "lora_B" in name:
             lora = True
+        # if lora naming convention, convert to traditional
         if "base_model.model." in name:
             name = name.replace("base_model.model.", "", 1)
         if "default." in name:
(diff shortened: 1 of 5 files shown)


plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py
```diff
@@ -278,28 +286,8 @@ def __init__(
...
-            num_experts=self.num_experts,
-            fan_out=1,
-            grouped_in=True,
-            dtype=dtype,
-            device=device,
-            lora_config=lora_config,
-        )
-        if mlp_arch == SCATTERMOE_SPEC_HAS_GATE:
-            self.w3 = ScatteredExperts(
+        if not lora_config or ("input_linear" in target_modules):
```
fabianlim, line 289: we shouldnt have hardcodes here

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py
```diff
@@ -457,36 +468,39 @@ def forward(self, hidden_states: torch.Tensor):
...
-            sorted_scattered_idxs,
-            padded_block_idxs,
-            expert_offsets,
-        )
-        out = self.activation(out)
-
-        # - if the arch has a seperate gate projection
-        if self.w3:
-            out *= self.w3(
+        if hasattr(self, "w1"):
```
fabianlim, line 471: shouldnt it be the case that we always have a `w1`?

willmj: With these limited changes, we are not supporting expert training for lora with ScatterMoE since inference cannot be run on the experts in base vLLM

plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
```diff
@@ -110,16 +112,30 @@ def save_fsdp_optimizer(
...
-    ckpt_model = os.path.join(output_dir, f"{FSDP_MODEL_NAME}_{MODEL_INDEX}")
-    os.makedirs(ckpt_model, exist_ok=True)
-    logger.info(f"Saving model to {ckpt_model}")
-    dcp.save(
-        state_dict={KEY_MODEL: model_state_dict},
-        storage_writer=dcp.FileSystemWriter(ckpt_model),
-        planner=DefaultSavePlanner(),
-    )
-    logger.info(f"Model saved to {ckpt_model}")
+    if lora_state_dict:
```
fabianlim, line 121: if len(lora_state_dict) > 0:
  ....

this might be more readable.

plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
```diff
@@ -540,14 +580,28 @@ def recover_safetensors_from_dcp(
     # get the state_dict
     state_dict = loader(checkpoint_dir)
 
+    new_state_dict = {}
+    lora = False
```
fabianlim, line 584: could this be simplified? or better commented.

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py
```diff
@@ -278,28 +273,8 @@ def __init__(
...
-            num_experts=self.num_experts,
-            fan_out=1,
-            grouped_in=True,
-            dtype=dtype,
-            device=device,
-            lora_config=lora_config,
-        )
-        if mlp_arch == SCATTERMOE_SPEC_HAS_GATE:
-            self.w3 = ScatteredExperts(
+        if not lora_config:
```
fabianlim, line 276: if you do this, then there is no module swap with `ScatteredExperts` , and then you do not have any kernels. So if you do not want to train the experts with lora, then all you need to do 

is to:
1. do not activate `lora_config` in `ScatteredExperts`
2. ensure that the `w1`, `w2`, `w3` have their `requires_grad==False`

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py
fabianlim: see above,

fabianlim (changes requested):
we need to be careful not to disable the ScatterMoE if you do not want to lora train the experts. Some changes are are required to fix that issue

plugins/accelerated-moe/src/fms_acceleration_moe/utils/checkpoint_utils.py
willmj: After some looking into this, I'd like to go with `if lora_state_dict` since dicts are falsy in Python and I think it is cleaner. If you feel strongly about using `len`, we can go for it though, just let me know

plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py
```diff
@@ -307,8 +300,11 @@ def __init__(
                 grouped_out=True,
                 dtype=dtype,
                 device=device,
-                lora_config=lora_config,
             )
+        if lora_config:
```
fabianlim, line 304: this can be considered a workaround, since we do not necessarily want to have requires_grad == False if `lora_config` is provided. In the future you would want to tune these also.

willmj: I'll add a comment, thanks

fabianlim (approved):

PR #142, opened 2025-04-29
fix: filter out router so `all-linear` can be tuned for LoRA target modules

willmj:
Goes with [#545](https://github.com/foundation-model-stack/fms-hf-tuning/pull/545) to allow all-linear for ScatterMoE. This PRs always filters out the `router` in the LoRA state dict, which should be reversed when expert tuning support for LoRA layers is added in, and has been noted as such. This will allow users to pass `all-linear` as target modules when using ScatterMoE

### DIFF
--- a/plugins/accelerated-moe/tox.ini
+++ b/plugins/accelerated-moe/tox.ini
@@ -4,6 +4,7 @@ envlist = py, lint
 [testenv]
 deps = 
     pytest>=7
+    importlib-metadata
     -e {toxinidir}
 skip_install = true
 commands = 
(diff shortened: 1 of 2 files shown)


plugins/accelerated-moe/tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py, lint
 [testenv]
 deps = 
     pytest>=7
+    importlib-metadata
```
anhuong, line 7: Error coming from linting failing to import from torch
```
.tox/py/lib/python3.9/site-packages/torch/distributed/elastic/rendezvous/__init__.py:142: in <module>
    from .registry import _register_default_handlers, _register_out_of_tree_handlers
.tox/py/lib/python3.9/site-packages/torch/distributed/elastic/rendezvous/registry.py:19: in <module>
    from importlib_metadata import entry_points
E   ModuleNotFoundError: No module named 'importlib_metadata'
```

Which looks like something that was added 7 months ago - https://github.com/pytorch/pytorch/blame/main/torch/distributed/elastic/rendezvous/registry.py#L18 

So not sure why this is coming up as a new error when it didn't fail [previously](https://github.com/foundation-model-stack/fms-acceleration/actions/runs/14514579135/job/40720653091). But this should not affect the image build since we run on python 3.12 and this import only happens on python < 3.10.

anhuong (approved):

PR #143, opened 2025-05-12
feat: Support granite 4 preview architecture for MoE kernels, EP, and fast kernels

kmehant:
Extend support to GraniteMoeHybridForCausalLM for all the fast moe (fast kernels, and EP) features and padding free.

## Summary of changes

- [x] extend support to GraniteMoeHybridForCausalLM
- [x] update a100_80gb_moe.csv |  updated based on new bench results and new GraniteMoeHybridForCausalLM model
- [x] update a100_80gb_moe.csv | bench for mixtral
- [x] update [requirements_moe.txt](https://github.com/foundation-model-stack/fms-acceleration/blob/main/scripts/benchmarks/refs/requirements_moe.txt)
- [x] regression test for the above runs.
- [x] report parity charts
- [x] report outliers
- [x] raise issue for failed runs in the new benchmark (see: https://github.com/foundation-model-stack/fms-acceleration/issues/147)
- [x] raise issue for loss regression (see: https://github.com/foundation-model-stack/fms-acceleration/issues/147)

## Final reg plots that includes past models

| memory increased             |  loss | throughput increased |
:-------------------------:|:-------------------------:|:-------------------------:
compare-mem_nvidia_mem_reserved  |  compare-train_loss | compare-train_tokens_per_second

## Outliers

3 classes of outliers can be identified
1. increased throughput
2. increased memory consumption
3. increased loss

issue: https://github.com/foundation-model-stack/fms-acceleration/issues/147

### Loss regression

Models `ibm-granite/granite-3.0-3b-a800m-instruct` and `ibm-research/moe-7b-1b-active-shared-experts` all padding-free runs regressed from previous bench loss showing larger losses than previous bench loss. However, its not clear if it has to do with padding-free since other models in the benchmark set didn't regress with padding free on.

Screenshot 2025-06-13 at 12 17 34 PM

### All outliers

Screenshot 2025-06-13 at 12 21 36 PM

### Reason: OOM
Screenshot 2025-06-13 at 2 23 20 PM

## Granite 4 preview acceleration over baselines (mamba kernels + accelerations added as part of this PR.)
Screenshot 2025-06-13 at 3 03 02 PM


plugins/accelerated-peft/requirements.txt
```diff
@@ -5,10 +5,10 @@
...
 
 # bitsandbytes for the BNB plugin
-# - lower bound is because bnb is missing quant_state
-# - upper bound is because of segmentation faults
-# see https://github.com/foundation-model-stack/fms-acceleration/issues/17
-bitsandbytes >=0.41,<=0.43.3
+# versions above 0.45.1 to support torch 2.6
+# exact version is used since upper bound is not known
+
+bitsandbytes == 0.45.1
```
fabianlim, line 11: in this case isnt it better to just lower bound? and if so this line `exact version is used since upper bound is not known` is not needed

tox.ini
```diff
@@ -34,6 +34,9 @@ commands =
     # some models need this for tokenizers
     pip install protobuf
 
+    # install mamba kernels to benchmark mamba based models
```
fabianlim, line 37: this is a very troublesome package. it is better to only install this if an environment variable is spcified. look at the tox documentation, the env can be accessed using `{env:NAME}`

as for the NAME. we need to decide on a good one

fabianlim (changes requested):
better not to have mamba installs by default

plugins/accelerated-peft/requirements.txt
kmehant: Fixed @fabianlim

tox.ini
kmehant: Thanks Fabian, have added a version for this, please review!

fabianlim (approved):
Im ok with the changes. Just wondering what caused the loss regressionl

kmehant:
> Just wondering what caused the loss regressionl

It was happening only on old models and padding free setting, also not for newer models with padding free on, so needs investigation, we will check it based on available cycles and update it on the attached issue.

PR #146, opened 2025-06-05
fix: Update QuantLinear import to GPTQLoraLinear

Luka-D:
I am trying to resolve [this issue](https://github.com/foundation-model-stack/fms-hf-tuning/pull/556) with dependabot updating to a new version of the `peft` library. The update has not been merged as it is encountering the following error:

```
from peft.tuners.lora.gptq import QuantLinear as LoraLinearGPTQ
ImportError: cannot import name 'QuantLinear' from 'peft.tuners.lora.gptq' (/home/tuning/.local/lib/python3.12/site-packages/peft/tuners/lora/gptq.py)
```

Turns out this is because `QuantLinear` has been renamed to `GPTQLoraLinear` in the newer version of `peft`. This can be seen in [this commit](https://github.com/huggingface/peft/commit/8c8bf8f1c870ab5be9c5da6c3da740cb5155c214).

So I have changed the import statement to import `GPTQLoraLinear` and modified the code not to use `LoraLinearGPTQ` anymore, as I felt it was pretty similar to the base class name which we could use instead.

Please let me know your thoughts.

### DIFF
--- a/plugins/accelerated-peft/pyproject.toml
+++ b/plugins/accelerated-peft/pyproject.toml
@@ -23,9 +23,6 @@ classifiers=[
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
 ]
-dependencies = [
-  "peft>=0.15",
-]
 
 [project.optional-dependencies]
 flash-attn = ["flash-attn"]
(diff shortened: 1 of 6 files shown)


Luka-D:
Thanks Fabian, I have pushed a new commit specifying `"peft>=0.15"` in the dependencies for foak and accelerate-peft

PR #162, opened 2025-11-18
feat: ODM without categories

romitjain:
1. Added a new `DatasetAutoCategorizer` class that creates an embedding out of text data and clusters it using KMeans (GPU accelerated)
2. The above is automatically triggered if ODM is used with a single category (i.e a dataset dict with a single key/value)
3. To run the tests: `python -m pytest tests/test_auto_categorization.py`

### DIFF
--- a/plugins/online-data-mixing/src/fms_acceleration_odm/odm/auto_categorizer.py
+++ b/plugins/online-data-mixing/src/fms_acceleration_odm/odm/auto_categorizer.py
@@ -25,6 +25,7 @@ import copy
 import math
 
 # Third Party
+import torch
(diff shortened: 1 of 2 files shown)


romitjain:
@kmehant the builds failed due to an OSError

```
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```

Do you know how to fix it?

kmehant:
Please amend or remove this commit - 0adf9c030180cc5f5151291146c41e1735a14eee

romitjain:
@kmehant I will figure out why the linting github actions are not running, but meanwhile, you can review the logic and PR

